PR #26, opened 2015-01-21
Add date to transactions

Gataniel:


app/models/transaction.rb
```diff
@@ -31,6 +31,7 @@ class Transaction < ActiveRecord::Base
   validates :bank_account, presence: true
   validates :transaction_type, inclusion: { in: TRANSACTION_TYPES, allow_blank: true }
 
+  after_create :set_date
```
AlexanderUlitin: @Gataniel I meant in task that we need only to add this column to UI table (not db table).

PR #27, opened 2015-01-23
Bank transfer

Gataniel:


app/controllers/transactions_controller.rb
```diff
@@ -7,6 +7,12 @@ def create
     @transaction.save
   end
 
+  def create_transfer
+    @transfer = Transfer.new(transaction_params)
+    @transaction = Transaction.new(transaction_params)
```
AlexanderUlitin: @Gataniel why do you need this Transaction Instance in this action? I think you can operate with Transfer instance, what you think?

app/views/transactions/_transactions.html.slim
```diff
@@ -1,4 +1,6 @@
 .row
   = render 'transactions/form', transaction: Transaction.new
+  = render 'transactions/transfer_form', transaction: Transaction.new
```
AlexanderUlitin: @Gataniel why you use Transaction instance here? Looks like this form for creating Transfer?

lib/transfer.rb
```diff
@@ -0,0 +1,72 @@
...
+          Category::CATEGORY_BANK_INCOME_PARAMS).id,
+        transaction_type: 'Receipt')
+
+      if (out_transaction.errors.any? && inc_transaction.errors.any?)
+        transaction.errors[:base] << out_transaction.errors.full_messages
+        transaction.errors[:base] << inc_transaction.errors.full_messages
+      end
+    else
+      errors.each do |key, value|
+        transaction.errors.add(key, value)
```
AlexanderUlitin: @Gataniel I think you don't need to copy errors to passed Transaction instance. You can operate with Transfer instance.

lib/transfer.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+
+    [inc_transaction, out_transaction]
+  end
+
+  private
+
+    def transfer_amount
+      if !bank_account_id.empty? &&
+          BankAccount.find(bank_account_id).balance.to_f < amount.to_f + comission.to_f
```
AlexanderUlitin: @Gataniel move `BankAccount.find(bank_account_id)` to method with local caching.

lib/transfer.rb
```diff
@@ -0,0 +1,72 @@
+class Transfer
```
shir: @Gataniel put it in `app/models`. `lib` folder is for library files. But this is business logic.

app/controllers/transactions_controller.rb
shir: @Gataniel @AlexanderUlitin yes, you don't need transaction here. Use transactions (if you need) only through transfer.

app/controllers/transactions_controller.rb
```diff
@@ -7,6 +7,12 @@ def create
...
+    @out_transaction, @inc_transaction  = @transfer.save(@transaction)
```
shir: @Gataniel don't return transactions on save. User something like `@transfer.in_transaction` to get transactions after save. `save` method should return `true`/`false`

shir: @Gataniel add errors handling.

app/controllers/transactions_controller.rb
```diff
@@ -7,6 +7,12 @@ def create
     @transaction.save
   end
 
+  def create_transfer
+    @transfer = Transfer.new(transaction_params)
```
shir: @Gataniel use `transfer_params` here.

app/models/category.rb
```diff
@@ -13,7 +13,19 @@
 class Category < ActiveRecord::Base
   CATEGORY_INCOME = 'Income'
   CATEGORY_EXPENSE = 'Expense'
+  CATEGORY_BANK_INCOME  = 'Bank receipt'
```
shir: @Gataniel why `BANK`? transfer is not only for bank transfers.

app/models/transaction.rb
```diff
@@ -13,10 +13,14 @@
 
 class Transaction < ActiveRecord::Base
   CURRENCIES = %w(USD RUB)
-  TRANSACTION_TYPES = %w(Residue)
+  TRANSACTION_TYPES = %w(Residue Receipt Transfer)
```
shir: @Gataniel you don't need to add additional transaction types. You already have category to check transaction type.

app/models/transaction.rb
```diff
@@ -13,10 +13,14 @@
...
+
+  attr_accessor :comission
 
   belongs_to :category, inverse_of: :transactions
   belongs_to :bank_account, inverse_of: :transactions
+  belongs_to :bank_account, inverse_of: :transactions
+  belongs_to :reference, class_name: 'BankAccount', inverse_of: :transactions
```
shir: @Gataniel it's not inverse of `transactions`. this association doesn't has any inverse association

shir: @Gataniel why do you need this field?

app/views/transactions/_btn_block.html.slim
```diff
@@ -0,0 +1,13 @@
+.col-sm-4.btn-group.pull-right
```
shir: @Gataniel it's not a good idea to use `col-sm` with `pull-right`. And move `col-sm` outside of partial.

app/views/transactions/_transfer_form.html.slim
```diff
@@ -0,0 +1,16 @@
+= simple_form_for(transaction, url: transactions_transfer_path, remote: true, html: {id: 'new_transfer_form'} ) do |f|
```
shir: @Gataniel this form should be based on Transfer model, not on Transaction. And Transfer model creates required transaction objects.

app/views/transactions/create.js.erb
```diff
@@ -1,7 +1,9 @@
-<% if @transaction.persisted? %>
-  $("<%=j render('transactions/transaction', transaction: @transaction) %>").hide().prependTo(".transactions").fadeIn(1000);
-  $("#new_transaction").replaceWith("<%=j render('transactions/form', transaction: Transaction.new(category_id: params[:transaction][:fixed_category_id]), fixed_category_id: params[:transaction][:fixed_category_id]) %>");
-  scrollTo($("#<%= dom_id(@transaction) %>"));
-<% else %>
-  $("#new_transaction").replaceWith("<%=j render('transactions/form', transaction: @transaction, fixed_category_id: params[:transaction][:fixed_category_id]) %>");
+<% if @transaction %>
```
shir: @Gataniel why do you need this? This view should be rendered without `@transaction` variable. If it's rendered some way then something goes wrong.

config/application.rb
```diff
@@ -27,6 +27,8 @@ class Application < Rails::Application
     # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
     # config.i18n.default_locale = :de
 
+    config.autoload_paths << Rails.root.join('lib')
```
shir: @Gataniel remove it. Lib folder should not be autoloaded.

config/initializers/simple_form.rb
```diff
@@ -60,7 +60,7 @@
   # Method used to tidy up errors. Specify any Rails Array method.
   # :first lists the first message for each field.
   # Use :to_sentence to list all errors for each field.
-  # config.error_method = :first
+  config.error_method = :to_sentence
```
shir: @Gataniel I think you don't need this too.

config/routes.rb
```diff
@@ -16,5 +16,6 @@
   resources :bank_accounts, except: [:show, :index]
   resources :categories
   resources :transactions,  only: [:create, :edit, :update, :destroy]
+  post 'transactions/transfer', to: 'transactions#create_transfer'
```
shir: @Gataniel 
```rb
resources :transactions,  only: [:create, :edit, :update, :destroy] do
  post :transfer, action: :create_transfer
end
```
http://guides.rubyonrails.org/routing.html

lib/transfer.rb
```diff
@@ -0,0 +1,72 @@
...
+  def initialize(attributes = {})
+    attributes.each do |name, value|
+      send("#{name}=", value)
+    end
+  end
+
+  def save(transaction)
+    if valid?
+      out_transaction = Transaction.create(
+        amount: amount.to_f + comission.to_f, bank_account_id: bank_account_id,
```
shir: @Gataniel make `amount` and `commission` as money fields. Don't use `to_f` for money. It's possible to lose data with it.

lib/transfer.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      inc_transaction = Transaction.create(
+        amount: amount, bank_account_id: reference_id,
+        reference_id: bank_account_id, comment: comment + "\nCommission: " + comission,
+        category_id: Category.find_or_create_by(
+          Category::CATEGORY_BANK_INCOME_PARAMS).id,
+        transaction_type: 'Receipt')
+
+      if (out_transaction.errors.any? && inc_transaction.errors.any?)
+        transaction.errors[:base] << out_transaction.errors.full_messages
```
shir: @Gataniel don't use full messages here. Copy errors for specified fields.

app/models/transfer.rb
```diff
@@ -0,0 +1,107 @@
...
+        category_id: Category.find_or_create_by(
+          Category::CATEGORY_BANK_EXPENSE_PARAMS).id)
+
+      @inc_transaction = Transaction.create(
+        amount_cents: @amount_cents,
+        bank_account_id: @reference_id, comment: form_comment(@comment),
+        category_id: Category.find_or_create_by(
+          Category::CATEGORY_BANK_INCOME_PARAMS).id)
+
+      if (@out_transaction.errors.any? || @inc_transaction.errors.any?)
```
shir: @Gataniel it's better use:
```rb
@out_transaction = Transaction.new(...)
@inc_transaction = Transaction.new(...)

if !@out_transaction.save || !@inc_transaction.save
   ... # TODO: collect errors
   return false
else
  ...
  return true
```
because save may be false without producing any errors.

app/models/transfer.rb
```diff
@@ -0,0 +1,107 @@
...
+  def money_amount
+    Money.new(@amount_cents, bank_account.currency)
+  end
+
+  def money_comission
+    Money.new(@comission_cents, bank_account.currency)
+  end
+
+  def amount=(value)
+    if value
```
shir: @Gataniel if not value then set `@amount_cents` and `@amount` to nil.

app/models/transfer.rb
```diff
@@ -0,0 +1,107 @@
...
+      @amount_cents = value.to_d * 100
+      @amount = value
+    end
+  end
+
+  def comission=(value)
+    if value
```
shir: @Gataniel same here.

PR #32, opened 2015-02-02
Fix rubles formatting

Gataniel:


config/initializers/money.rb
```diff
@@ -65,3 +65,11 @@
   # config.no_cents_if_whole = nil
   # config.symbol = nil
 end
+
+module MoneyRails
```
shir: @Gataniel don't modify MoneyRails. Update configuration for Ruble currency instead.

PR #33, opened 2015-02-03
Add filter on transactions

Gataniel:


app/models/transaction.rb
```diff
@@ -51,4 +53,29 @@ def recalculate_amount
...
+
+  def self.period(period)
+    case period
+    when "current_month"
+      where("transactions.created_at >= ?", Time.now.beginning_of_month)
+    when "last_3_months"
+      where("transactions.created_at >= ?", (Time.now - 3.months).beginning_of_day)
+    when "last_month"
+      last_month_begins = Time.now.beginning_of_month - 1.months
+      where("transactions.created_at >= ? AND transactions.created_at <= ?", last_month_begins,
```
shir: @Gataniel you may use [`BETWEEN`](http://www.postgresql.org/docs/9.4/static/functions-comparison.html) operator

config/application.rb
```diff
@@ -1,5 +1,5 @@
 require File.expand_path('../boot', __FILE__)
-
+ENV['RANSACK_FORM_BUILDER'] = '::SimpleForm::FormBuilder'
```
shir: @Gataniel try to put it in `/config/initializers/ransack`

Gemfile
```diff
@@ -18,6 +18,8 @@ gem 'uglifier', '>= 1.3.0'
 # Asset libraries
 gem 'jquery-rails'
 gem 'bootstrap-sass'
+gem 'momentjs-rails'
+gem 'bootstrap-daterangepicker-rails'
```
shir: @Gataniel looks like this gems are not used. Remove them.

config/application.rb
Gataniel: @shir didn't work

PR #34, opened 2015-02-03
Show currencies flow

Gataniel:


app/models/transaction.rb
```diff
@@ -25,8 +27,10 @@ class Transaction < ActiveRecord::Base
   delegate :income?, :expense?, to: :category, allow_nil: true
 
   default_scope { order(created_at: :desc) }
+  scope :rub, -> { joins(:bank_account).where('bank_accounts.currency' => "RUB") }
+  scope :usd, -> { joins(:bank_account).where('bank_accounts.currency' => "USD") }
```
shir: @Gataniel instead use one scope `by_currency(currency)`

app/views/transactions/_transactions.html.slim
```diff
@@ -1,4 +1,10 @@
 .row
   = render 'transactions/form', transaction: Transaction.new
+hr
+  h4 Filters:
+.row
+  = render 'transactions/filter'
+.row
+  = render 'transactions/flow'
```
shir: @Gataniel should be shown only if filter is applied.

PR #36, opened 2015-02-04
Add transaction table sort

Gataniel:


app/models/transaction.rb
```diff
@@ -37,6 +37,10 @@ class Transaction < ActiveRecord::Base
   after_save :recalculate_amount
   after_destroy :recalculate_amount
 
+  def date
+    created_at.strftime("%Y-%m-%d")
```
shir: @Gataniel instead use `I18n.l` method in view and put default format in `config/locale/..`

PR #37, opened 2015-02-04
Add soft-delete

Gataniel:


app/models/bank_account.rb
```diff
@@ -30,6 +32,7 @@ class BankAccount < ActiveRecord::Base
   validates :currency, presence: true, inclusion: { in: CURRENCIES, message: "%{value} is not a valid currency" }
 
   after_create :set_initial_residue
+  before_restore "recalculate_amount!(with_deleted = true)"
```
shir: @Gataniel same here about code in string.

shir: @Gataniel Not sure why do you need this callback.

app/models/transaction.rb
```diff
@@ -33,6 +35,7 @@ class Transaction < ActiveRecord::Base
   before_save :check_negative
   after_save :recalculate_amount
   after_destroy :recalculate_amount
+  before_restore "recalculate_amount(with_deleted = true)"
```
shir: @Gataniel same here.

spec/features/soft_delete.rb
```diff
@@ -0,0 +1,34 @@
+require 'spec_helper'
```
shir: @Gataniel it's not a feature. Put it in specified model specs.

app/models/transaction.rb
```diff
@@ -36,6 +41,9 @@ class Transaction < ActiveRecord::Base
   before_save :check_negative
   after_save :recalculate_amount
   after_destroy :recalculate_amount
+  before_restore do
```
shir: @Gataniel could you describe why do you need it? Looks like it works wrong.

spec/models/bank_account_spec.rb
```diff
@@ -11,4 +11,27 @@
     it { should validate_presence_of(:currency) }
     it { should ensure_inclusion_of(:currency).in_array(%w(USD RUB)) }
   end
+
+  context "soft delete" do
```
shir: @Gataniel it's not `context` it's `describe`

spec/models/transaction_spec.rb
```diff
@@ -12,6 +12,29 @@
     it { should validate_presence_of(:bank_account) }
   end
 
+  context 'soft delete' do
```
shir: @Gataniel same here.

app/models/transaction.rb
Gataniel: @shir If we dont put it here when we restore transaction, bank account balance won't be recalculated and his sum will be like there is no such transaction. Will be outdated

app/models/transaction.rb
```diff
@@ -26,7 +28,10 @@ class Transaction < ActiveRecord::Base
   delegate :income?, :expense?, to: :category, allow_nil: true
 
   default_scope { order(created_at: :desc) }
-  scope :by_currency, ->(currency) { joins(:bank_account).where('bank_accounts.currency' => currency) }
+  scope :by_currency, ->(currency) { joins("INNER JOIN bank_accounts bank_account_transactions
+      ON bank_account_transactions.id = transactions.bank_account_id
+      AND bank_account_transactions.deleted_at IS NULL
+      AND bank_accounts.currency = '#{currency}'") }
```
shir: @Gataniel escape `currency`. Not sure why do you move it to `joins` instead `where`

PR #41, opened 2015-02-06
Bug long input 500 error

Gataniel:


app/models/transfer.rb
```diff
@@ -6,7 +6,8 @@ class Transfer
     :inc_transaction, :out_transaction
 
   validates :amount, presence: :true, numericality: { greater_than: 0 }
-  validates :comission, numericality: { greater_than: 0 }
+  validates :comission, numericality: { greater_than: 0 }, length: { maximum: 255 }
```
AlexanderUlitin: @Gataniel commission can me blank.

app/models/transaction.rb
```diff
@@ -28,7 +28,7 @@ class Transaction < ActiveRecord::Base
   default_scope { order(created_at: :desc) }
   scope :by_currency, ->(currency) { joins(:bank_account).where('bank_accounts.currency' => currency) }
 
-  validates :amount, presence: true
+  validates :amount, presence: true, length: { maximum: 255 }
```
AlexanderUlitin: @Gataniel I suggest to use lower max_length, cos it's a integer in db... Let's use `length: { maximum: 10 }`

app/models/transfer.rb
AlexanderUlitin: @Gataniel same here, let's set max length as 10 symbols.

PR #46, opened 2015-02-06
Set db max value for amount

Gataniel:


app/models/transaction.rb
```diff
@@ -28,7 +28,7 @@ class Transaction < ActiveRecord::Base
   default_scope { order(created_at: :desc) }
   scope :by_currency, ->(currency) { joins(:bank_account).where('bank_accounts.currency' => currency) }
 
-  validates :amount, presence: true, length: { maximum: 20 }
+  validates :amount, presence: true
```
shir: @Gataniel why didn't you add same validation here?

app/models/bank_account.rb
```diff
@@ -26,7 +26,8 @@ class BankAccount < ActiveRecord::Base
   monetize :residue_cents, with_model_currency: :currency
 
   validates :name,     presence: true
-  validates :balance,  presence: true
+  validates :balance,  presence: true, numericality: {
+    less_than_or_equal_to: 21_474_836.47 }
```
shir: @Gataniel put this value in constant or in config.

spec/models/transaction_spec.rb
```diff
@@ -11,22 +11,6 @@
     it { should validate_presence_of(:category)     }
     it { should validate_presence_of(:bank_account) }
 
-    context "validate length of :amount" do
```
shir: @Gataniel instead this add spec for max value validation.

spec/features/organizations_list_spec.rb
```diff
@@ -24,8 +24,6 @@
         expect(subject).to have_selector('td',
           text: organization.owners.map(&:email).join(', '))
         expect(subject).to have_link('Show')
-        expect(subject).to have_link('Edit')
```
shir: @Gataniel why did you remove this tests? There is nothing in this task that should change it.

PR #47, opened 2015-02-06
Add custom date transactions filter

Gataniel:


app/assets/javascripts/transactions.js
```diff
@@ -1,4 +1,6 @@
 $(function () {
+  show_hide_period_additional_input();
```
shir: @Gataniel you should call this function if appropriate elements are on page.

PR #51, opened 2015-02-09
Fix exchange rate input values

Gataniel:


app/models/transfer.rb
```diff
@@ -85,6 +84,13 @@ def comission=(value)
     end
   end
 
+  def currency_mismatch?
+    @from_currency = bank_account.try(:currency)
+    @to_currency   = BankAccount.find_by_id(reference_id).try(:currency)
+
+    @from_currency != @to_currency ? true : false
```
shir: @Gataniel why can't you just return `@from_currency != @to_currency`? And why do you use class variables here?

app/models/transfer.rb
```diff
@@ -99,18 +105,13 @@ def transfer_account
     end
 
     def bank_account
-     @bank_account ||= BankAccount.find(@bank_account_id)
+     @bank_account ||= BankAccount.find_by_id(@bank_account_id)
```
shir: @Gataniel use `bind_by(id: ...)`. And use accessor methods instead class variables: `bank_account_id` instead `@bank_accont_id`

PR #55, opened 2015-02-10
Finding system cats and add specs

Gataniel:


app/controllers/categories_controller.rb
```diff
@@ -37,14 +37,19 @@ def destroy
   end
 
   def show
-    @q = @category.transactions.ransack(params[:q])
+    @q = if @category.system?
+      current_organization.transactions.where(category_id: @category.id).ransack(params[:q])
+    else
+      @category.transactions.ransack(params[:q])
```
shir: @Gataniel why can't you use same query here as in previous condition?

app/controllers/categories_controller.rb
```diff
@@ -37,14 +37,19 @@ def destroy
...
+    end
     @transactions = @q.result.page(params[:page])
   end
 
   private
     # Use callbacks to share common setup or constraints between actions.
     def set_category
-      @category = current_organization.categories.find(params[:id])
+      category = Category.find(params[:id])
+      @category = category.system? ? category : current_organization.categories.find(params[:id])
```
shir: @Gataniel it's better to make
```rb
@category = Category.for_organization(current_organizaiton).find(params[:id])

class Category
  scope :for_organization, ->(organization){ TODO: select system or by organization id }
end
```

app/models/transfer.rb
```diff
@@ -51,6 +51,10 @@ def save
     end
   end
 
+  def save!
+    save ? self : ActiveRecord::RecordInvalid
```
shir: @Gataniel `save!` should raise exception in case of error not return.

app/controllers/categories_controller.rb
Gataniel: @shir system category belongs to no organization. It will not be found

app/models/category.rb
```diff
@@ -38,6 +38,8 @@ class Category < ActiveRecord::Base
 
   scope :incomes,  -> { where(type: CATEGORY_INCOME)  }
   scope :expenses, -> { where(type: CATEGORY_EXPENSE) }
+  scope :for_organization, ->(organization) {
+    where("system = true or organization_id = ?", organization.id) }
```
shir: @Gataniel rewrite it:
```
where("categories.system = ? OR catgories.organization_id =?", true, organization.id)
```

app/models/transfer.rb
```diff
@@ -51,6 +51,10 @@ def save
     end
   end
 
+  def save!
+    save ? self : (raise self.inspect)
```
shir: @Gataniel 
```
def save!
  raise self.inspect unless save
end
```

PR #61, opened 2015-02-10
Fix long inputs issue

Gataniel:


app/models/bank_account.rb
```diff
@@ -47,7 +47,7 @@ def set_initial_residue
     transactions.create(amount_cents: residue_cents, transaction_type: 'Residue') if residue > 0
   end
 
-  def to_s
-    "#{name} (#{Money::Currency.new(currency).symbol})"
+  def to_label
```
shir: @Gataniel why did you rename method? Leave it as is.

Instead truncating whole string, truncate only name, so currency symbol will be still visible.

app/models/category.rb
```diff
@@ -56,4 +56,7 @@ def expense?
     type == CATEGORY_EXPENSE
   end
 
+  def to_label
```
shir: @Gataniel rename to `to_s`

app/assets/stylesheets/transactions.css.scss
```diff
@@ -1,3 +1,7 @@
 #new_transfer_form {
   display: none;
 }
+
+.word_break {
```
shir: @Gataniel use dash `-` instead underscore `_` for css class names.

PR #63, opened 2015-02-10
app/views/organizations/

Gataniel:


app/views/organizations/edit.html.slim
```diff
@@ -2,7 +2,7 @@ h1 Editing organization
 
 == render 'form'
 
-= link_to 'Show', @organization
+=' link_to 'Show', @organization
```
shir: @Gataniel update `slim` gem and use `=>` instead `='`

spec/features/organizations_list_spec.rb
```diff
@@ -24,8 +24,6 @@
         expect(subject).to have_selector('td',
           text: organization.owners.map(&:email).join(', '))
         expect(subject).to have_link('Show')
-        expect(subject).to have_link('Edit')
```
shir: @Gataniel why is it here?

PR #69, opened 2015-02-12
Disable password on profile update

Gataniel:


app/controllers/user/registrations_controller.rb
```diff
@@ -0,0 +1,27 @@
+class User::RegistrationsController < Devise::RegistrationsController
+
+  def update_profile
+    self.resource = resource_class.to_adapter.get!(send(:"current_#{resource_name}").to_key)
```
shir: @Gataniel why do complicated? I guess devise already has method to get current resource.
You should take a look at this wiki https://github.com/plataformatec/devise/wiki/How-To:-Allow-users-to-edit-their-account-without-providing-a-password

PR #75, opened 2015-02-17
Add menu options

Gataniel:


app/views/home/_settings.html.slim
```diff
@@ -0,0 +1,7 @@
+.list-group
+  li
+    = link_to 'Bank accounts', organization_path(current_organization), class: active_sett('organizations')
+  li
+    = link_to 'Categories', categories_path, class: active_sett('categories')
+  li
```
AlexanderUlitin: @Gataniel try to use `active_link_to` https://github.com/comfy/active_link_to?

PR #76, opened 2015-02-17
Add drag&drop sorting to accounts | Add specs

Gataniel:


app/assets/javascripts/bank_accounts.js
```diff
@@ -0,0 +1,29 @@
...
+      update: function(e, ui) {
+        var itemId   = ui.item.data("item-id");
+        var position = ui.item.index();
+        return $.ajax({
+          type: "POST",
+          url: $('#bank_accounts_table').attr('data-url'),
+          dataType: "json",
+          data: {
+            id: itemId,
+            position: position
```
AlexanderUlitin: @Gataniel why do you need this variables here? I think it will be simpler just to use hash like:
```
data: {
           id: ui.item.data("item-id"),
           position: ui.item.index()
}
```

AlexanderUlitin:
@shir do you think we can merge this PR?

app/assets/javascripts/bank_accounts.js
```diff
@@ -0,0 +1,27 @@
+$(document).on("ready page:load", function () {
+  if ($("#bank_accounts_table").length > 0) {
```
shir: @Gataniel why don't you use any exist library for drag and drop?

app/views/bank_accounts/_bank_accounts.html.slim
```diff
@@ -9,7 +9,7 @@ h1 Listing bank accounts
         th Balance
     tbody
       - bank_accounts.each do |bank_account|
-        tr
+        tr data-item-id="#{bank_account.id}" class="item"
```
shir: @Gataniel why do you use quotes here?

config/routes.rb
```diff
@@ -16,6 +16,7 @@
   end
   resources :bank_accounts, except: :index do
     put :hide, on: :member
+    post :sort, on: :collection
```
shir: @Gataniel use put or patch method for it.

db/seeds.rb
```diff
@@ -5,3 +5,38 @@
 #
 #   cities = City.create([{ name: 'Chicago' }, { name: 'Copenhagen' }])
 #   Mayor.create(name: 'Emanuel', city: cities.first)
+User.destroy_all
```
shir: @Gataniel why it's all here? I think we already told about it that there should not be test data in seeds.

app/assets/javascripts/bank_accounts.js
Gataniel: @shir Don't know any, this one I used on kpkrf project

shir: @Gataniel ok

PR #80, opened 2015-02-18
Add input mansks for amount and custom date

Gataniel:


app/assets/javascripts/application.js
```diff
@@ -15,6 +15,7 @@
 //= require bootstrap
 //= require moment
 //= require daterangepicker
+//= require jquery.mask.min
```
AlexanderUlitin: @Gataniel please use not minified code for libs. It's hard to debug it in case issues and anyway we use assets pipeline for minifying.

PR #91, opened 2015-02-24
Include edge dates

Gataniel:


app/models/transaction.rb
```diff
@@ -89,7 +89,7 @@ def self.period(period)
   end
 
   def self.date_from(from)
-    from = DateTime.parse(from) rescue nil
+    from = DateTime.parse(from).beginning_of_day rescue nil
```
AlexanderUlitin: @Gataniel let's cover it in specs.

PR #103, opened 2015-03-03
Currency lists

Gataniel:


app/models/bank_account.rb
```diff
@@ -64,4 +62,18 @@ def set_initial_residue
   def to_s
     "#{name.truncate(30)} (#{Money::Currency.new(currency).symbol})"
   end
+
+  class << self
+    def grouped_by_currency(currency)
+      currencies = Currency.ordered(currency)
+
+      grouped = []
```
shir: @Gataniel use ruby-way for such things
```
includes(:currency).all.index_by(:currency).sort_by{ |c, a| c.position }
```

app/models/transfer.rb
```diff
@@ -132,7 +132,9 @@ def estimate_amount(out)
       estimated_amount
     end
 
-    def parse_amount
+    def parse_masked
```
shir: @Gataniel look like it will be not needed after another task will be done.

app/models/currency.rb
```diff
@@ -0,0 +1,8 @@
+class Currency
+  def self.ordered(default_currency = "USD")
+    currencies = Dictionaries.to_h[:currencies].sort
+    i = currencies.index(default_currency)
+    currencies[0], currencies[i] = currencies[i], currencies[0]
```
shir: @Gataniel you should not swap. You should just put default currency at first place.

app/views/home/_sidebar.html.slim
```diff
@@ -2,14 +2,16 @@
...
+  - bank_accounts = current_organization.bank_accounts.visible.positioned
+
   - if bank_accounts.present?
-    table.table.table-hover
-      - bank_accounts.positioned.each do |account|
-        tr class=(account.balance_cents > 0 ? 'success' : 'danger') id="bank_account_#{account.id}"
-          td= link_to account.to_s, root_path(q: { bank_account_id_eq: account.id })
-          td.bank-amount
-            = money_with_symbol account.balance
+    - Currency.ordered(current_organization.default_currency).each do |curr|
```
shir: @Gataniel not sure why do you use here `Currency.ordered` if you already have method to get accounts sorted by currency. Use it here.

app/models/bank_account.rb
Gataniel: @shir didn't understand how to do it via index_by. 
`BankAccount.all.index_by(:currency) ArgumentError: wrong number of arguments (1 for 0)
`
did it the next way:
```rb
  class << self
    def grouped_by_currency(def_currency)
      currencies = Currency.ordered(def_currency)

      all.group_by(&:currency).sort_by do |ba|
        currencies.index(ba.first)
      end
    end
  end
```

Gataniel: `BankAccount.all.index_by(&:currency)` returns only 1 bank account for every currency. I think it rewrites keys values

PR #109, opened 2015-03-05
add-clear-filter

Gataniel:


app/assets/javascripts/transactions.js
```diff
@@ -26,6 +26,15 @@ $(function () {
...
   });
 
+  $(document).on('click', '#clear', function(e) {
+    e.preventDefault();
+
+    $('#q_amount_eq').val("");
+    $('#q_comment_cont').val("");
+    $('#q_category_id_eq').val("");
+    $('#q_bank_account_id_eq').val("");
+    $('#q_period').val("");
```
AlexanderUlitin: @Gataniel please use this method:

```
$(this).closest('form').(':input')
  .removeAttr('checked')
  .removeAttr('selected')
  .not(':button, :submit, :reset, :hidden, :radio, :checkbox')
  .val('');
```

app/assets/javascripts/transactions.js
```diff
@@ -26,6 +26,15 @@ $(function () {
     e.stopPropagation();
   });
 
+  $(document).on('click', '#clear', function(e) {
```
AlexanderUlitin: @Gataniel let's use some class name, like `clear-form`, so we will able to use it in a few forms on one page in future.

PR #111, opened 2015-03-06
Show filtered total in default currency

Gataniel:


app/models/transaction.rb
```diff
@@ -51,15 +51,34 @@ class Transaction < ActiveRecord::Base
...
-    ]
+  class << self
+    def flow_ordered(def_currency)
+      currencies = Currency.ordered(def_currency)
+
+      curr_flow = {}
+      currencies.each do |curr|
+        curr_flow[curr] = {}
+
+        transactions = by_currency(curr)
```
shir: @Gataniel I think it will be better to make this via sql with one query. It will work much faster.

shir: @Gataniel and create separate object/structure to store flow.

app/views/home/_flow.html.slim
```diff
@@ -1,24 +1,25 @@
...
+    .row
       | Total:
-    .col-sm-2 class=(@rub_flow > 0 ? 'bg-success' : 'bg-danger')
-      = money_with_symbol(@rub_flow)
-    .col-sm-2 class=(@usd_flow > 0 ? 'bg-success' : 'bg-danger')
-      = money_with_symbol(@usd_flow)
+  - def_curr = current_organization.default_currency
+  - @curr_flow.each_pair do |curr, calc|
+    .col-sm-2
+      .row class=(calc[:inc] > 0 ? 'bg-success' : 'bg-danger')
```
shir: @Gataniel put class calculation in helper.

spec/models/transaction_spec.rb
```diff
@@ -214,4 +214,29 @@
...
+    let(:def_trans)   { org.transactions.by_currency(def_curr) }
+
+    let(:slave_inc) { Money.new(slave_trans.incomes.sum(:amount_cents), slave_curr)}
+    let(:slave_exp) { Money.new(slave_trans.expenses.sum(:amount_cents), slave_curr)}
+    let(:def_inc) { Money.new(def_trans.incomes.sum(:amount_cents), def_curr)}
+    let(:def_exp) { Money.new(def_trans.expenses.sum(:amount_cents), def_curr)}
+
+    subject { org.transactions.flow_ordered(org.default_currency) }
+
+    it "contain right values" do
```
shir: @Gataniel it's bad description.

app/helpers/money_helper.rb
```diff
@@ -1,11 +1,11 @@
...
 
-  def money_sign(money)
-    money > 0 ? 'success' : 'danger'
-  end
-
   def money_with_symbol(money)
     humanized_money_with_symbol(money, symbol_after_without_space: true)
   end
+
+  def bg(amount)
```
AlexanderUlitin: @Gataniel please name this method more clear, like `amount_bg_class` or something similar.

app/views/home/_flow.html.slim
```diff
@@ -1,24 +1,25 @@
 #flow.h4
-  .row
-    .col-sm-2
+  .col-sm-2
```
shir: @Gataniel you should use table here. Use table to show table data

PR #116, opened 2015-03-10
Config accounts colors

Gataniel:


config/config.yml.sample
```diff
@@ -1,5 +1,14 @@
 defaults: &defaults
-
+  account_colors:
+    negative:
+      css_class: 'negative'
```
AlexanderUlitin: @Gataniel I think we don't need this class variable here... Just store color value.

app/helpers/bank_accounts_helper.rb
```diff
@@ -0,0 +1,11 @@
+module BankAccountsHelper
+  def colorize(amount)
+    if amount > 0
+      AppConfig.account_colors.positive.css_class
```
AlexanderUlitin: @Gataniel don't use class from config, let's think that class always match variable name (positive, negative, e.t.c)

config/config.yml.sample
```diff
@@ -1,5 +1,14 @@
 defaults: &defaults
-
+  account_colors:
+    negative:
```
AlexanderUlitin: @Gataniel let's make it like this:

```
account_colors:
  negative: '#fa8072'
  positive: '#a0db8e'
  empty: '#fffff'
```

PR #119, opened 2015-03-11
Fix filter long input issue

Gataniel:


app/views/transactions/_filter.html.slim
```diff
@@ -1,8 +1,9 @@
 h4 Filters:
 
 = search_form_for @q, url: root_path do |f|
-  = f.input :amount_eq, label: false, input_html: { class: 'amount' }
-  = f.input :comment_cont, label: false
+  = f.input :amount_eq, label: false, input_html: { class: 'amount',
+    maxlength: 20 }
```
AlexanderUlitin: @Gataniel I think it can be on previous line

PR #130, opened 2015-03-12
Save rate to db

Gataniel:


lib/russian_central_bank_safe.rb
```diff
@@ -0,0 +1,24 @@
...
+  def update_rates(date = Date.today)
+      @mutex.synchronize{
+        begin
+          update_parsed_rates exchange_rates(date)
+          @rates_updated_at = Time.now
+          @rates_updated_on = date
+          @rates
+        rescue Exception => e
+          begin
+            last_db_rates = ExchangeRate.last
```
shir: @Gataniel not sure why do you need this part?

lib/russian_central_bank_safe.rb
```diff
@@ -0,0 +1,24 @@
...
+            @rates_updated_at = last_db_rates.updated_from_bank_at
+            @rates_updated_on = date
+            @rates = last_db_rates.rates
+          rescue Exception => e
+            rates = YAML.load_file(Rails.root.join('db', 'seeds', 'rates.yml'))
```
shir: @Gataniel and this

spec/models/exchange_rate_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'spec_helper'
+
+describe ExchangeRate do
+  context 'validation' do
+    it { should validate_presence_of(:rates) }
```
shir: @Gataniel use `expect` syntax:
```rb
it { is_expected.to ... }
```

lib/russian_central_bank_safe.rb
Gataniel: @shir We try to update rates from bank
If HTTP 302 is returned then we try to take values from our DB.
But when server starts there is no connection to DB when initializing money.rb and setting default bank and rates, in that case we parse values from the yml file

PR #132, opened 2015-03-16
Colorize amounts

Gataniel:


app/views/transactions/_transaction.html.slim
```diff
@@ -1,12 +1,12 @@
-tr.transaction id=dom_id(transaction) class=(transaction.income? ? 'success' : 'danger') data-edit-url=edit_transaction_path(transaction)
+tr.transaction id=dom_id(transaction) class=colorize_amount(transaction.amount) data-edit-url=edit_transaction_path(transaction)
   td.col-sm-2 = money_with_symbol transaction.amount
 
   td.col-sm-3.word-break
-    = link_to transaction.category.name,
+    = link_to_if transaction.category, transaction.category.name,
```
AlexanderUlitin: @Gataniel can you please double-check that it will work? I think there is no sense to use `link_to_if` in line `link_to_if transaction.category ... if transaction.category`...

app/views/transactions/_transaction.html.slim
```diff
@@ -1,12 +1,12 @@
...
       root_path(q: (params[:q] || {}).merge(category_id_eq: transaction.category.id)),
       data: { stop_propagation: true } if transaction.category
   td.col-sm-3.word-break
-    = link_to transaction.bank_account.to_s,
+    = link_to_if transaction.bank_account, transaction.bank_account.to_s,
```
AlexanderUlitin: @Gataniel same here.

PR #134, opened 2015-03-17
Show sidebar menu on cat new

Gataniel:


spec/support/shared_examples/sidebar_menu_activatable.rb
```diff
@@ -0,0 +1,5 @@
+shared_examples_for "activatable" do |activated|
+  it 'is actived now' do
+    expect(page).to have_css('.active', activated)
```
AlexanderUlitin: @Gataniel are you sure thst it works properly? Maybe it should be something like `have_css('.active', text: activated)`?

AlexanderUlitin: @Gataniel also let's add within here, so we will search only in some specific area, not in whole page.

PR #139, opened 2015-03-18
Fix filter totals

Gataniel:


spec/support/shared_examples/filterable.rb
```diff
@@ -59,4 +59,6 @@ def expect_currencies(rub, usd)
       expect(page).to_not have_css(to_css_id(wrong_item))
     end
   end
+
+  # it_behaves_like "has flow"
```
AlexanderUlitin: @Gataniel please uncomment it or remove if you don't need it here.

PR #140, opened 2015-03-19
Create transaction only for current organization

Gataniel:


spec/controllers/transactions_controller_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  let!(:ba)   { create :bank_account, organization: org }
+  let!(:ba2)  { create :bank_account, organization: org2 }
+  let!(:cat)  { create :category, organization: org }
+  let!(:cat2) { create :category, organization: org2 }
+
+  before do
+    sign_in user
+  end
+
+  describe 'POST #create' do
```
AlexanderUlitin: @Gataniel let's add case when we create Transaction successfully.

spec/controllers/transactions_controller_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    context 'when category and bank from not current organization' do
+      subject { post :create, format: :js,
+        transaction: { amount: 100, category_id: cat2.id, bank_account: ba2.id } }
+      it_behaves_like "don't create model instance", Transaction
+    end
+
+  end
+
+  describe 'create_transfer' do
```
AlexanderUlitin: @Gataniel let's add case when we create Transfer successfully.

spec/support/shared_examples/dont_create_model_instance.rb
```diff
@@ -0,0 +1,5 @@
+shared_examples_for "don't create model instance" do |model|
+  it 'doesnt change count' do
+    expect{ subject }.to_not change{ model.count }
```
AlexanderUlitin: @Gataniel I'm not sure if it's make sense to move this one `expcet` to shared examples... @shir what do you think?

PR #143, opened 2015-03-23
Invitations index

Gataniel:


app/controllers/invitations_controller.rb
```diff
@@ -1,10 +1,17 @@
 class InvitationsController < ApplicationController
+  layout 'settings'
   skip_filter :authenticate_user!, only: :accept
-  before_filter :find_invitation, only: :accept
+  before_action :find_active_invitation, only: :accept
+  before_action :find_invitation, only: [:destroy, :resend]
 
   def new
+    authorize :invitation, :new?
```
shir: @Gataniel you don't need to add `new?` if action and policy has same name.

app/controllers/invitations_controller.rb
```diff
@@ -1,10 +1,17 @@
...
     @invitation = Invitation.new
-    authorize @invitation, :create?
+  end
+
+  def index
+    authorize :invitation, :index?
```
shir: @Gataniel same here

app/controllers/users/registrations_controller.rb
```diff
@@ -8,7 +8,7 @@ def create_user_from_invitation
     @user.email = @invitation.email
     if @user.save
       sign_in @user
-      redirect_to accept_invitation_path(token: @invitation.token)
+      redirect_to accept_invitation_path(@invitation, token: @invitation.token)
```
shir: @Gataniel why do you pass both `id` and `token`?

app/policies/invitation_policy.rb
```diff
@@ -1,5 +1,21 @@
...
+  def new?
+    member.owner_or_admin?
+  end
+
+  def index?
+    member.owner_or_admin?
+  end
+
   def create?
     (member.admin? && record.role != 'owner') || member.owner?
```
shir: @Gataniel put it function.

app/views/home/_settings.html.slim
```diff
@@ -9,3 +9,5 @@
       active: [['categories']]
     = active_link_to 'Members', members_path, class: 'list-group-item',
       active: [['members'],['index']]
+    = active_link_to 'Invitations', invitations_path, class: 'list-group-item',
+      active: [['invitations']] if current_member.owner_or_admin?
```
shir: @Gataniel is policy here `... if policy(:invitation).index?`

app/views/invitation_mailer/new_invitation.html.slim
```diff
@@ -1,4 +1,4 @@
 = "You are invited to #{h @organization.name} as #{h @invitation.role}"
 
 = "\n\nTo accept, please click the following link"
-= link_to 'Accept', accept_invitation_url(token: @invitation.token)
+= link_to 'Accept', accept_invitation_url(@invitation, token: @invitation.token)
```
shir: @Gataniel why do you pass both `id` and `token`?

app/views/invitation_mailer/new_invitation.text.erb
```diff
@@ -1,3 +1,3 @@
 <%= "You are invited to #{h @organization.name} as #{h @invitation.role}" %>
 
-To accept, please click the following link <%= accept_invitation_url(token: @invitation.token) %>
+To accept, please click the following link <%= accept_invitation_url(@invitation, token: @invitation.token) %>
```
shir: @Gataniel why do you pass both `id` and `token`?

app/controllers/users/registrations_controller.rb
Gataniel: I did following in routes.rb
```  resources :invitations, only: [:index, :new, :create, :destroy] do
    member do
      get :accept
      get :resend
    end
  end
```
`accept_invitation_path	GET	/invitations/:id/accept(.:format)	invitations#accept`
shoud I change it to collection route?

shir: @Gataniel you should not use `id` in invitation route, so user can't find foreign invitations just by changing `id`. You should always use token. So in this case you may just pass token as id:
```rb
accept_invitation_path(id: @invitation.token)
```
and in controller
```rb
def find_invitation
  Invitation.find_by!(token: params[:id])
end
```

Or declare a new route with token:
```rb
get `/invitation/:token/accept` => `invitations#accept`
```

app/controllers/invitations_controller.rb
```diff
@@ -34,11 +60,15 @@ def accept
 
   private
 
-  def find_invitation
+  def find_active_invitation
     @invitation = Invitation.active.find_by(token: params[:token])
```
shir: @yunixon you should search for invitation only within the current organization

app/controllers/invitations_controller.rb
```diff
@@ -34,11 +60,15 @@ def accept
...
     redirect_to root_path, alert: 'Bad invitation token' unless @invitation
   end
 
+  def find_invitation
+    @invitation = Invitation.find_by!(token: params[:token])
```
shir: @yunixon you should search for invitation only within the current organization

yunixon: @shir current_organization is nil at this place because new user not signed in yet

app/controllers/invitations_controller.rb
yunixon: @shir current_organization is nil at this place because new user not signed in yet

shir: @yunixon ok. But you still should fix for `find_invitation`

PR #145, opened 2015-03-24
Update project

Gataniel:


.ruby-version
```diff
@@ -1 +1 @@
-2.1.1
+2.2.0
```
shir: @Gataniel please update to 2.2.1. 2.2.0 has serious bugs.

app/views/devise/sessions/new.html.slim
```diff
@@ -1,12 +1,10 @@
...
-  .col-sm-6
-    = simple_form_for(resource, as: resource_name, url: session_path(resource_name)) do |f|
-      .form-inputs
-        = f.input :email, :required => false, :autofocus => true
-        = f.input :password, :required => false
-        = f.input :remember_me, :as => :boolean if devise_mapping.rememberable?
-      .form-actions
-        = f.button :submit, "Sign in", class: 'btn-primary'
-    = render "devise/shared/links"
+.row.col-sm-6
```
AlexanderUlitin: @Gataniel `.col-sm-6` should be inside `.row`.

config/initializers/simple_form.rb
```diff
@@ -66,7 +72,7 @@
   config.error_notification_tag = :div
 
   # CSS class to add for error notification helper.
-  config.error_notification_class = 'alert alert-error'
+  config.error_notification_class = 'error_notification'
```
AlexanderUlitin: @Gataniel why you use another class for errors?

spec/features/organizations_list_spec.rb
```diff
@@ -9,19 +9,17 @@
     visit organizations_path
   end
 
-  subject { page }
```
AlexanderUlitin: @Gataniel why you did this changes? I don't see any difference here.

config/initializers/simple_form.rb
Gataniel: It is updated Simple Form. Should I leave it as default or change to old alert alert-error

AlexanderUlitin: @Gataniel it depends on styles that we use in forms. As I know we use bootstrap and classes `alert alert-error` will show red colored alerts. Please check it and use styles that will works from Twitter Bootstrap.

AlexanderUlitin: @Gataniel it depends on styles that we use in forms. As I know we use bootstrap and classes `alert alert-error` will show red colored alerts. Please check it and use styles that will works from Twitter Bootstrap.

AlexanderUlitin: @Gataniel it depends on styles that we use in forms. As I know we use bootstrap and classes `alert alert-error` will show red colored alerts. Please check it and use styles that will works from Twitter Bootstrap.

Gataniel: @AlexanderUlitin  Actually I don't see any difference...Everything seems to work fine

PR #146, opened 2015-03-25
Update js for selection highlightening | add spec

Gataniel:


app/views/transactions/_btn_block.html.slim
```diff
@@ -1,12 +1,20 @@
...
   ul.dropdown-menu
-    li
-      = link_to 'Transfer', '#', remote: true, id: 'new_transfer_btn'
-      = link_to 'Transaction', '#', remote: true, id: 'new_transaction_btn'
+    - if selected == 'Transaction'
+      li
+        = link_to 'Transfer', '#', remote: true, id: 'new_transfer_btn'
+      li class='active'
+        = link_to 'Transaction', '#', remote: true, id: 'new_transaction_btn'
+    - else
```
AlexanderUlitin: @Gataniel why not add duplicated code here? like this:

```
li class=(selected == 'Transfer' ? 'active' : '')
  = link_to 'Transfer', '#', remote: true, id: 'new_transfer_btn'
li class=(selected == 'Transaction' ? 'active' : '')
  = link_to 'Transaction', '#', remote: true, id: 'new_transaction_btn'
```

PR #156, opened 2015-03-27
Add customer settings and soft delete

Gataniel:


app/models/customer.rb
```diff
@@ -4,20 +4,32 @@
...
+  # gem 'paranoia' doesn't run validations on restore
+  before_restore :name_uniqueness
 
   def to_s
     name.truncate(30)
   end
+
+  private
+    def name_uniqueness
+      errors.add(:name, 'has been taken', strict: true) if Customer.find_by(
```
shir: @Gataniel why don't you use uniqueness validation here?

Gataniel: @shir I dont quite understand. What do you mean? Can I somehow exmplicitly run validate_uniqueness_of? Didn't find something like that.. Can run record.validate_presense_of(:attr) etc, but seems to me there is no record.validate_uniqueness_of(:attr)

shir: @Gataniel I think you should just run `valid?`. You should run all validation or no one. Why only one?

PR #161, opened 2015-04-06
Add customer new for select2

Gataniel:


app/controllers/customers_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    if @customer.save
+      redirect_to customers_path, notice: 'Customer was successfully created.'
+    else
+      render action: 'new'
+    end
+  end
+
+  def autocomplete
+    @customers = current_organization.customers.pluck(:id, :name)
```
shir: @Gataniel add search by provided term. `Select2` may pass `term` to request to make search by

app/views/transactions/_form.html.slim
```diff
@@ -13,8 +13,9 @@
       - else
         = f.association :category, as: :grouped_select, collection: current_organization.categories.grouped_by_type, group_method: :last, value_method: :id, prompt: 'Category', label: false
     .col-md-3.col-sm-6.col-xs-6
-      = f.input :customer_id, as: :select, include_blank: 'Customer',
-        label: false, collection: current_organization.customers
+      = f.input :customer_id, label: false,
+        input_html: { data: { url: customers_path } }
+      = f.input :customer_name, label: false, as: :hidden
```
shir: @Gataniel don't add hidden field. Pass name as data attribute to customer_id field.

app/views/transactions/_form.html.slim
```diff
@@ -13,8 +13,9 @@
       - else
         = f.association :category, as: :grouped_select, collection: current_organization.categories.grouped_by_type, group_method: :last, value_method: :id, prompt: 'Category', label: false
     .col-md-3.col-sm-6.col-xs-6
-      = f.input :customer_id, as: :select, include_blank: 'Customer',
-        label: false, collection: current_organization.customers
+      = f.input :customer_id, label: false,
+        input_html: { data: { url: customers_path } }
```
shir: @Gataniel I think you should use url for autocomplete here. (see my comment above)

app/controllers/customers_controller.rb
Gataniel: @shir What is supposed to be in `term`? What kind of conditions? Something like
{name: customer_name}, so that we could 
``` @customers = current_organization.customers.where(name: customer_name) ? But will return only 1 customer

shir: @Gataniel there will be text that user has entered in select2. So you should select all customers which name starts with this term. Or include. So it be something like `where("name like ?", "%term%")`.

app/views/transactions/_form.html.slim
shir: @Gataniel Oj. I see..I think you should use only `name` field here. And on server side find client by name or create a new one.

PR #165, opened 2015-04-13
Fix create destroy transaction when only one

Gataniel:


app/views/transactions/destroy.js.erb
```diff
@@ -1,5 +1,11 @@
-$("#<%= dom_id(@transaction) %>").remove();
-$("#<%= dom_id(@transaction, 'edit_row') %>").remove();
+// 2 = transaction + transaction edit form
```
AlexanderUlitin: @Gataniel what this commented line means? Please clarify.

AlexanderUlitin: @Gataniel I see, it's explanation of 2 lines... Please write more human-readable comment, like '// Need to check if there are 2 lines, then reload page'

app/views/transactions/destroy.js.erb
```diff
@@ -1,5 +1,11 @@
...
+if ($('.transactions tbody tr').size() == 2) {
+  location.reload();
```
AlexanderUlitin: @Gataniel please clarify why we need page reloading here?

PR #166, opened 2015-04-13
Separate recently added transactions

Gataniel:


app/models/member.rb
```diff
@@ -2,12 +2,13 @@
...
-#  created_at      :datetime
-#  updated_at      :datetime
-#  role            :string           not null
+#  id                        :integer          not null, primary key
+#  user_id                   :integer          not null
+#  organization_id           :integer          not null
+#  created_at                :datetime
+#  updated_at                :datetime
+#  role                      :string           not null
+#  root_page_last_visited_at :datetime
```
AlexanderUlitin: @Gataniel let's remove this field to just `last_viewed_at`, without root_page?

app/views/transactions/create.js.erb
```diff
@@ -1,10 +1,23 @@
...
-  $("#new_transaction").replaceWith("<%=j render('transactions/form', transaction: Transaction.new(category_id: params[:transaction][:fixed_category_id]), fixed_category_id: params[:transaction][:fixed_category_id]) %>");
-  scrollTo($("#<%= dom_id(@transaction) %>"));
-  $("#sidebar").replaceWith("<%= j render('home/sidebar') %>");
-  $("#total_balance").replaceWith("<%= j render('shared/layout/total_balance') %>");
+  if ($('.transactions').size() == 0) {
+    location.reload();
+  }
+  else {
+    $("<%=j render('transactions/transaction', transaction: @transaction) %>").hide().prependTo(".transactions").fadeIn(1000);
+    if (document.URL == "<%= root_url %>" && $('.transactions tbody tr td[colspan]').size() == 0) {
```
AlexanderUlitin: @Gataniel I think we don't need this feature (when I see line under my new transaction). We need this feature only for new transactions if I have not seen them before, but when I create my transaction - I know about it and it should not be marked as unread. I suggest to use more deep colors for unread transactions, it will be easier to implement, all you need is just check if `transaction.created_at > current_member.last_viewed_at` in `app/views/transactions/_transaction.html.slim` file.

app/views/transactions/_transaction.html.slim
```diff
@@ -1,4 +1,4 @@
-tr.transaction id=dom_id(transaction) class=colorize_amount(transaction.amount) data-edit-url=edit_transaction_path(transaction)
+tr.transaction id=dom_id(transaction) class=colorize_amount(transaction.amount) data-edit-url=edit_transaction_path(transaction) class=colorize_transaction(transaction)
```
AlexanderUlitin: @Gataniel you set `class` 2 times in this line, it's not correct.

AlexanderUlitin: @Gataniel please don't use `colorize_amount` here, use `colorize_transaction` helper only and set color for positive and negative transactions in this helper too.

app/helpers/layout_helper.rb
```diff
@@ -2,4 +2,16 @@ module LayoutHelper
   def active_org(org)
     (org.id == current_organization.id ? 'active' : '')
   end
+
+  def colorize_transaction(transaction)
+    css_class = colorize_amount(transaction.amount)
+
+    if params[:controller] != 'transactions' && current_member.last_visited_at
+      if transaction.created_at > current_member.last_visited_at
+        css_class += ' new_transaction'
```
AlexanderUlitin: @Gataniel how about using more deep color instead of opacity? I would suggest to use color that I see when I hover on Transaction line.

app/controllers/application_controller.rb
```diff
@@ -44,4 +44,8 @@ def configure_permitted_parameters
     devise_parameter_sanitizer.for(:account_update) << [:full_name, profile_attributes: [:phone_number, :avatar]]
     devise_parameter_sanitizer.for(:sign_up) << [:full_name]
   end
+
+  def update_last_viewed_at
+    current_member.update_attribute(:last_visited_at, Time.now)
```
shir: @Gataniel don't use `update_attribute`, it skips validations. Use just `update`

PR #167, opened 2015-04-14
Fix creating first transaction js

Gataniel:


app/controllers/transactions_controller.rb
```diff
@@ -3,12 +3,14 @@ class TransactionsController < ApplicationController
   before_action :set_transaction,  only: [:edit, :update, :destroy]
 
   def create
+    @q = Transaction.ransack()
```
shir: @Gataniel why do you need this? I don't see it's used.

app/views/transactions/_transactions_table.html.slim
```diff
@@ -15,7 +15,7 @@
     = paginate transactions, remote: true
   .row.col-sm-12
     = render 'home/flow' if params[:q]
-- elsif params[:q].blank?
+- elsif params[:q].blank? && search == false
```
shir: @Gataniel I think it will be better to use variable like `hide_search` and then use it `if local_assigns[:hide_search]`. In this case you may to not set this variable in render there it's not used.

app/views/transactions/create.js.erb
```diff
@@ -1,15 +1,20 @@
 <% if @transaction.persisted? %>
   if ($('.transactions').size() == 0) {
-    // need to reload because there is no ransack initialized & no transactions table
-    location.reload();
+    $('.transactions_list').
+      html("<%=j render partial: 'transactions/transactions_table',
+        locals: { transactions:
+          Kaminari.paginate_array([@transaction]).page(params[:page]).per(50),
```
shir: @Gataniel why it's here? I don't understand why do you need to render single transaction as table. let's discuss it. It looks wrong.

app/views/transactions/destroy.js.erb
```diff
@@ -1,4 +1,14 @@
-$('.transactions_list').html("<%=j render partial: 'transactions/transactions_table', locals: { transactions: @transactions } %>");
+$("#<%= dom_id(@transaction) %>").remove();
+$("#<%= dom_id(@transaction, 'edit_row') %>").remove();
+
+if ($('.transactions tbody tr').size() == 0) {
+  if (document.URL != "<%= root_url %>") {
```
shir: @Gataniel Instead checking url just check if search form already exists. And just pass this condition as parameter

PR #168, opened 2015-04-15
Fix tests

Gataniel:


spec/support/shared_examples/paginateable.rb
```diff
@@ -0,0 +1,47 @@
+shared_examples_for "paginateable" do
+  describe 'page' do
+    before do
+      visit list_page
+    end
+
+    it "lists first page items" do
+      list.first(paginated).each do |item|
+        within list_class do
+          expect(page).to have_css("#{item_id_prefix}#{item.id}")
```
AlexanderUlitin: @Gataniel why not just use `dom_id(item)` here?

PR #177, opened 2015-07-14
fix exchange rate

yunixon:


app/models/transfer.rb
```diff
@@ -127,12 +127,14 @@ def form_comment(comment)
     def estimate_amount(out)
       estimated_amount = out ? (amount_cents + comission_cents) : amount_cents
       if currency_mismatch?
-        Money.add_rate(from_currency, to_currency, exchange_rate.to_d)
+        rate = Money.default_bank.get_rate(from_currency, to_currency)
```
shir: @yunixon why do you need this?

app/models/exchange_rate.rb
```diff
@@ -15,8 +15,9 @@ class ExchangeRate < ActiveRecord::Base
 
   class << self
     def update_rates
+      Money.default_bank.update_rates
```
shir: @yunixon please, write specs for this function to make sure all works ok.

app/models/transfer.rb
yunixon: Because "Money.add_rate" set global rate for this pare currencies, i temporary save previous value to local var and then return value: Money.default_bank.add_rate(from_currency, to_currency, rate)

shir: @yunixon ok

PR #179, opened 2015-07-15
customers crud, specs, fix pagination for categories

yunixon:


app/views/customers/index.json.jbuilder
```diff
@@ -0,0 +1,4 @@
+json.array!(@customers) do |customer|
```
shir: @yunixon we don't need it.

spec/features/customers_page_spec.rb
```diff
@@ -0,0 +1,73 @@
+require 'spec_helper'
```
shir: @yunixon add spec for customers index page on which check if customers from current organization are shown, customers from other organizations are not shown, and deleted customers are not shown.

PR #181, opened 2015-07-16
Add select2 new customer

yunixon:


app/controllers/transactions_controller.rb
```diff
@@ -40,7 +40,7 @@ def check_relation_to_curr_org(trans)
     trans.reference_id =
       curr_bank_accounts.find_by_id(tparams[:reference_id]).try(:id) if tparams[:reference_id]
     trans.customer_id =
-      current_organization.customers.find_by_id(tparams[:customer_id]).try(:id) if trans == @transaction
+      current_organization.customers.find_by_name(tparams[:customer_name]).try(:name) if trans == @transaction
```
shir: @yunixon you use name for id field. Looks like something is wrong.

app/views/transactions/_form.html.slim
```diff
@@ -13,8 +13,9 @@
           collection: current_organization.categories.grouped_by_type,
           group_method: :last, prompt: 'Category', label: false, value_method: :id
     .col-md-3.col-sm-6.col-xs-6
-      = f.input :customer_id, as: :select, include_blank: 'Customer',
-        label: false, collection: current_organization.customers
+      = f.input :customer_name, label: false,
+        input_html: { data: { url: autocomplete_customers_path,
+          placeholder: f.object.customer_name || 'Customer'} }
```
shir: @yunixon you should not use customer name as placeholder. You should set it as predefined value. There is an option in select2 for it.

PR #184, opened 2015-07-21
Add statistics: charts with tabs

yunixon:


app/assets/javascripts/statistics.js
```diff
@@ -0,0 +1,88 @@
+$(function () {
+  getStatistics('current_month');
```
shir: @yunixon this function will be called on each page not depending if there is graphic or not.

app/assets/javascripts/statistics.js
```diff
@@ -0,0 +1,88 @@
...
+});
+
+function drawChart(response, css_id, title, period) {
+  if (response == null ) {
+    $(css_id).removeAttr('id').addClass('alert alert-warning').html('No data');
+    return false;
+  };
+  var data = response.data;
+  var ids  = response.ids;
+  var pieData = google.visualization.arrayToDataTable(data);
```
shir: @yunixon where `google` come from? I don't see where you add this library.

app/models/organization.rb
```diff
@@ -45,4 +45,56 @@ def exchange_rates
...
+    end
+  end
+
+  def calc_total_for_customer(customers, selection, data, ids)
+    customers.each_pair do |id, name|
+      total_for_customer = 0
+      selection.each do |income|
+        total_for_customer += income[:total] if income[:customer_id] == id
+      end
+      ids  << id
```
shir: @yunixon it will be better to use hash instead two arrays. And don't provide this hash as parameter. Create and return it from function. Later if you need separate arrays for ids and data you may cal `hash.keys` and `hash.values`.

app/views/layouts/settings.html.slim
```diff
@@ -10,3 +10,5 @@ html
           = yield :sidebar
         .col-sm-9
           = yield
+    footer
+      = yield :google_charts_api
```
shir: @yunixon make it `yield :footer`. So there may be any data rendered, not only google chart api.

app/assets/javascripts/statistics.js
```diff
@@ -0,0 +1,88 @@
...
+    getStatistics('this_year');
+  });
+
+  $( "a[href$='all-time']" ).click(function() {
+    getStatistics('all_time');
+  });
+
+});
+
+function drawChart(response, css_id, title, period) {
```
shir: @yunixon it will be better if you put all this functions in one class.

app/views/statistics/index.html.slim
```diff
@@ -1,5 +1,13 @@
...
+  h1 Statistics
+
+= render 'periods_bar'
 
 = content_for :sidebar
   = render 'home/settings'
+
+= content_for :google_charts_api
+  = javascript_include_tag 'https://www.google.com/jsapi'
+  = javascript_include_tag 'statistics'
```
shir: @yunixon don't include statistics js separately. It should be included in `application.js`

app/assets/javascripts/statistics.js
```diff
@@ -0,0 +1,73 @@
+$(function () {
+  path = window.location.href.split('/');
+  if (path[path.length - 1] == 'statistics') {
```
shir: @yunixon instead just check that container for graphic exists

app/assets/javascripts/statistics.js
```diff
@@ -0,0 +1,73 @@
...
+    getStatistics('current-month');
+  };
+  $(document).on('click', '#periods_bar li a', function(e) {
+    getStatistics($(this).data('period'));
+  });
+});
+
+function drawChart(response, css_id, title, period) {
```
shir: @yunixon it will be better if you put all this functions in one class.

PR #189, opened 2015-07-23
fix transacs date and set start date

yunixon:


app/assets/javascripts/transactions.js
```diff
@@ -129,6 +129,7 @@ function addRemoveHints(fromCurr, toCurr) {
 function datepickerInit(selector) {
   selector.datepicker({
     format: 'dd/mm/yyyy',
+    startDate: '01.01.1900',
```
shir: @yunixon looks like you use wrong format.

app/views/transactions/_form.html.slim
```diff
@@ -24,9 +24,8 @@
         = f.input :bank_account, label: false, disabled: true
     .col-md-3.col-sm-6.col-xs-6
       = f.input :comment, label: false
-    - if params[:controller] == 'transactions'
-      .col-md-3.col-sm-6.col-xs-6
-        = f.input :created_at, as: :string, label: false, input_html: { class: 'datepicker',
-          value: f.object.created_at.try(:strftime,'%d/%m/%Y') }
+    .col-md-3.col-sm-6.col-xs-6
+      = f.input :created_at, as: :string, label: false, input_html: { class: 'datepicker',
```
shir: @yunixon don't use `created_at` field for date. Create a new field `date` and use it.

app/controllers/home_controller.rb
```diff
@@ -3,10 +3,10 @@ class HomeController < ApplicationController
   before_action :update_last_viewed_at
 
   def show
-    @q = current_organization.transactions.ransack(params[:q])
+    @q = current_organization.transactions.ordered.ransack(params[:q])
```
shir: @yunixon for ransack you should set default ordering another way:
```rb
@q = ...
@q.sorts = 'date asc' if @q.sorts.blank?
@transactions = ...
```
With this default ordered column will be shown in UI and allow user to set another order.

app/models/transaction.rb
```diff
@@ -40,7 +41,7 @@ def total
   delegate :currency, to: :bank_account, allow_nil: true
   delegate :income?, :expense?, to: :category, allow_nil: true
 
-  default_scope { order(created_at: :desc) }
+  scope :ordered, -> { order('created_at DESC') }
```
shir: @yunixon order by `date` field here.

shir: @yunixon looks like you will not need this line after changing ransack ordering.

app/models/transaction.rb
```diff
@@ -55,7 +56,8 @@ def total
   validates :category, presence: true, unless: :residue?
   validates :bank_account, presence: true
   validates :transaction_type, inclusion: { in: TRANSACTION_TYPES, allow_blank: true }
-  validates :created_at, presence: true, on: :update
+  validates :date, presence: true, on: :update
```
shir: @yunixon date should be filled on creation too. Just fill it with current date if it is blank on `before_validate` callback

app/models/transaction.rb
```diff
@@ -118,6 +120,10 @@ def check_negative
     nil
   end
 
+  def check_date
```
shir: @yunixon you don't need this.

app/models/transfer.rb
```diff
@@ -32,12 +32,14 @@ def save
       @out_transaction = Transaction.new(
         amount_cents: estimate_amount(out = true),
         bank_account_id: bank_account_id, comment: form_comment(comment),
+        date: date,
```
shir: @yunixon before filling date for transaction set it to some value if it blank, so both transaction will have same date.

app/views/transactions/_transaction.html.slim
```diff
@@ -14,4 +14,4 @@ tr.transaction id=dom_id(transaction) class=colorize_transaction(transaction) da
       root_path(q: (params[:q] || {}).merge(customer_id_eq: transaction.customer_id)),
       data: { stop_propagation: true } if transaction.customer_id
   td.col-sm-2.word-break.comment.hidden-xs = transaction.comment
-  td.col-sm-1= I18n.l transaction.created_at
+  td.col-sm-1= I18n.l transaction.date if transaction.date
```
shir: @yunixon transaction always have date.

db/migrate/20150723081851_add_date_to_transactions.rb
```diff
@@ -0,0 +1,15 @@
+class AddDateToTransactions < ActiveRecord::Migration
+  def up
+    add_column :transactions, :date, :datetime
+    Transaction.all.find_each do |t|
```
shir: @yunixon instead this you may to it with one query:
```
Transaction.update_all('date = created_at')
```

db/migrate/20150723081851_add_date_to_transactions.rb
```diff
@@ -0,0 +1,15 @@
+class AddDateToTransactions < ActiveRecord::Migration
+  def up
+    add_column :transactions, :date, :datetime
```
shir: @yunixon change field to not null after fill all data.

db/migrate/20150723081851_add_date_to_transactions.rb
```diff
@@ -0,0 +1,15 @@
...
+    Transaction.all.find_each do |t|
+      # especially without validations
+      t.created_at ? t.update_attribute(:date, t.created_at) : t.update_attribute(:date, Time.now)
+    end
+    add_index :transactions, :date
+  end
+
+  def down
```
shir: @yunixon instead making `up` and `down` use `reversible`
```rb
def change
  add_column ...
  reversible do |dir|
    dir.up do
      Transaction.update_all(...)
      change_column ..., null: false
    end
    add_index ...
end
```

spec/features/transactions_filter_spec.rb
```diff
@@ -254,13 +254,13 @@
 
     context 'when custom' do
       let!(:transaction)  { Timecop.travel(2013,12,12) {
-        create :transaction, bank_account: ba } }
+        create :transaction, bank_account: ba, date: Time.now } }
```
shir: @yunixon why do you will everywhere? Just change factory.

yunixon: Timecop.travel(1.month.ago) { create :transaction...} Not working with date: Time.now in factory

shir: @yunixon because you should use in factory: `date: { Time.now }`. Read about lazy attributes in factory girl. https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#lazy-attributes

db/migrate/20150723081851_add_date_to_transactions.rb
```diff
@@ -0,0 +1,14 @@
+class AddDateToTransactions < ActiveRecord::Migration
+  def change
+    add_column :transactions, :date, :datetime
+    reversible do |dir|
+      dir.up do
+        Transaction.with_deleted.where(date: nil).update_all('date = created_at')
+        change_column :transactions, :date, :datetime, null: false
+      end
+      dir.down do
```
shir: @yunixon you don't need `dir.down` if it's empty

db/migrate/20150723081851_add_date_to_transactions.rb
```diff
@@ -0,0 +1,14 @@
+class AddDateToTransactions < ActiveRecord::Migration
+  def change
+    add_column :transactions, :date, :datetime
+    reversible do |dir|
+      dir.up do
+        Transaction.with_deleted.where(date: nil).update_all('date = created_at')
```
shir: @yunixon you just added `date` column. It will be nil for all rows.

spec/features/transactions_filter_spec.rb
```diff
@@ -285,6 +285,10 @@
           click_on 'Search'
         end
 
+        after do
+          Timecop.travel(Time.now)
```
shir: @yunixon use `Timecop.return` instead.

db/migrate/20150723081851_add_date_to_transactions.rb
yunixon: What about production server? All transactions will have date Time.now

shir: @yunixon you just add column in this migration. How it may be filled? Not dependent on development on production. If you already run migration on production, then you should prepare new migration to fill data, or rollback this (and delete column) and run it again.

PR #203, opened 2015-07-30
fix sum table view, helper refactoring, css refactoring

yunixon:


app/assets/stylesheets/bank_accounts.css.scss
```diff
@@ -17,6 +17,14 @@
   background-color: $positive-color !important;
 }
 
+.positive.darken {
```
shir: @yunixon don't call it `darken`. Use meaning name. Something like it was before `new-transaction`. If you want to use style for several classes, then write them all here `.new-transasaction, .total`. Or use mixin.

app/helpers/money_helper.rb
```diff
@@ -2,13 +2,11 @@ module MoneyHelper
...
 
   def colorize_amount(amount)
-    if amount > 0
-      'positive'
-    elsif amount < 0
-      'negative'
-    else
-      'empty'
-    end
+    amount > 0 ? 'positive' : (amount < 0 ? 'negative' : 'empty')
```
shir: @yunixon don't use several trinary operation in one string. It's hard to read.

PR #206, opened 2015-08-04
fix table sum view

yunixon:
Update sum table data then transactions create/update/destroy and filter on

### DIFF
--- a/app/models/bank_account.rb
+++ b/app/models/bank_account.rb
@@ -43,6 +43,7 @@ class BankAccount < ActiveRecord::Base
       message: "%{value} is not a valid currency" }
 
   after_create :set_initial_residue
+  after_touch :recalculate_amount!
 
   def residue_cents
     @residue_cents ||= 0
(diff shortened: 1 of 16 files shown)


app/views/transactions/create_transfer.js.erb
```diff
@@ -7,6 +7,10 @@
   scrollTo($("#<%= dom_id(@out_transaction) %>"));
   $("#sidebar").replaceWith("<%= j render('home/sidebar') %>");
   $("#total_balance").replaceWith("<%= j render('shared/layout/total_balance') %>");
+
+  if ($("#flow").length) {
```
shir: @yunixon it's wrong fix

app/models/bank_account.rb
```diff
@@ -43,6 +43,7 @@ class BankAccount < ActiveRecord::Base
       message: "%{value} is not a valid currency" }
 
   after_create :set_initial_residue
+  after_touch :recalculate_amount!
```
shir: :+1:

PR #212, opened 2015-08-06
fix switch

yunixon:


app/controllers/application_controller.rb
```diff
@@ -11,6 +11,7 @@ class ApplicationController < ActionController::Base
   helper_method :current_member
 
   rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
+  rescue_from ActiveRecord::RecordNotFound, with: :record_not_found
```
shir: @yunixon not sure why do you need it? It's ok to show 404 on this error (it's default behaviour).

PR #228, opened 2015-08-14
fix ba view

yunixon:


app/views/home/_sidebar.html.slim
```diff
@@ -5,9 +5,8 @@
       - bank_accounts.grouped_by_currency(current_organization.default_currency).each do |by_curr|
         table.table.table-hover
           - by_curr.last.each do |account|
-            tr class=colorize_amount(account.balance_cents)
+            tr class=colorize_ba_amount(account.balance_cents) id="bank_account_#{account.id}"
```
shir: @yunixon use `dom_id` for id generation: `id=dom_id(account)`

PR #231, opened 2015-08-18
transfer refactoring

yunixon:
step 1

### DIFF
--- a/app/helpers/layout_helper.rb
+++ b/app/helpers/layout_helper.rb
@@ -4,9 +4,15 @@ module LayoutHelper
   end
 
   def colorize_transaction(transaction)
-    css_class = colorize_amount(transaction.amount)
-    if params[:controller] == 'home' && current_member
-      if current_member.last_visited_at
+    if transaction.category && transaction.category.name == 'Receipt'
+      css_class = 'transfer'
+    else
+      css_class = colorize_amount(transaction.amount)
+    end
+    if params[:controller] == 'home' && current_member && current_member.last_visited_at
+      if transaction.category && transaction.category.name == 'Receipt'
+        css_class += ' new-transfer' if transaction.created_at > current_member.last_visited_at
+      else
         css_class += ' new-transaction' if transaction.created_at > current_member.last_visited_at
       end
     end
--- a/app/helpers/layout_helper.rb
+++ b/app/helpers/layout_helper.rb
@@ -4,17 +4,9 @@ module LayoutHelper
   end
 
   def colorize_transaction(transaction)
-    if transaction.category && transaction.category.name == 'Receipt'
-      css_class = 'transfer'
-    else
-      css_class = colorize_amount(transaction.amount)
-    end
+    css_class = transaction.transfer? ?  'transfer' : colorize_amount(transaction.amount)
     if params[:controller] == 'home' && current_member && current_member.last_visited_at
-      if transaction.category && transaction.category.name == 'Receipt'
-        css_class += ' new-transfer' if transaction.created_at > current_member.last_visited_at
-      else
-        css_class += ' new-transaction' if transaction.created_at > current_member.last_visited_at
-      end
+      css_class += ' new-transaction' if transaction.created_at > current_member.last_visited_at
     end
     css_class
   end
--- a/app/models/transaction.rb
+++ b/app/models/transaction.rb
@@ -51,9 +51,7 @@ class Transaction < ActiveRecord::Base
       where('bank_account_transactions.currency' => currency) }
   scope :incomes,     -> { joins(:category).where('categories.type' => Category::CATEGORY_INCOME)}
   scope :expenses,    -> { joins(:category).where('categories.type' => Category::CATEGORY_EXPENSE)}
-  scope :without_out, -> { joins('LEFT OUTER JOIN categories category_transactions
-      ON category_transactions.id = transactions.category_id').
-      where('category_transactions.id IS NULL OR category_transactions.name != ?', Category::CATEGORY_TRANSFER_OUTCOME) }
+  scope :without_out, -> { where('category_id IS NULL OR category_id != ?', Category.transfer_out_id) }
 
   validates :amount, presence: true, numericality: { greater_than: 0,
     less_than_or_equal_to: Dictionaries.money_max }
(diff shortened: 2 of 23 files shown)


app/models/transaction.rb
```diff
@@ -62,6 +66,7 @@ def total
   before_validation :find_customer, if: Proc.new{ customer_name.present? && bank_account.present? }
   before_validation :set_date, if: Proc.new{ date.blank? }
   before_save :check_negative
+  before_destroy :destroy_transfer, if: Proc.new{ transfer_id.present? }
```
shir: @yunixon why don't you use `dependent: :destroy` option on association instead?

app/models/transfer.rb
```diff
@@ -124,7 +125,8 @@ def bank_account
   end
 
   def form_comment(comment)
-    comment.to_s + "\nComission: " + (comission.blank? ? "0" : comission.to_s)
+    rate = currency_mismatch? ? "\nRate: " + exchange_rate.to_s : ''
+    comment.to_s + "\nComission: " + (comission.blank? ? "0" : comission.to_s) + rate
```
shir: @yunixon don't add commission comment if commission is 0.

app/models/transaction.rb
```diff
@@ -34,6 +34,7 @@ def total
   belongs_to :category, inverse_of: :transactions
   belongs_to :bank_account, inverse_of: :transactions, touch: true
   belongs_to :customer, inverse_of: :transactions
+  belongs_to :transfer, class_name: 'Transaction'
```
shir: @yunixon also create reverse association

shir: @yunixon it's a bad name for this column. Rename it to `transfer_out`

app/models/transaction.rb
```diff
@@ -49,6 +50,9 @@ def total
       where('bank_account_transactions.currency' => currency) }
   scope :incomes,     -> { joins(:category).where('categories.type' => Category::CATEGORY_INCOME)}
   scope :expenses,    -> { joins(:category).where('categories.type' => Category::CATEGORY_EXPENSE)}
+  scope :without_transfers, -> { joins('LEFT OUTER JOIN categories category_transactions
```
shir: @yunixon it's a bad name for scope. You load only without out transfer transactions.

shir: @yunixon why do you use `left outer join` and check id category is null? As I remember category is persisted.

And instead make join and load whole category table you may only load id of transfer category and filter by id

app/views/transactions/create_transfer.js.erb
```diff
@@ -1,13 +1,7 @@
...
-  $("<%=j render('transactions/transaction', transaction: @out_transaction) %>").hide().prependTo(".transactions").fadeIn(1000);
   $("<%=j render('transactions/transaction', transaction: @inc_transaction) %>").hide().prependTo(".transactions").fadeIn(1000);
-  var c_out = $("#<%= dom_id(@out_transaction) %>").css('backgroundColor');
   var c_inc = $("#<%= dom_id(@inc_transaction) %>").css('backgroundColor');
-  $("#<%= dom_id(@out_transaction) %>").addClass('new-transaction');
-  $("#<%= dom_id(@inc_transaction) %>").addClass('new-transaction');
-  $("#<%= dom_id(@out_transaction) %>").animate({
-    backgroundColor: c_out,
-  }, 1000 );
+  $("#<%= dom_id(@inc_transaction) %>").addClass('new-transfer');
```
shir: @yunixon use same `new-transaction` class. Why do you need new class here?

app/models/transaction.rb
yunixon: Category is null in Residue transactions, and its terrible :(

shir: @yunixon ok. When just select by category id.

app/views/transactions/create_transfer.js.erb
yunixon: Because color is yellow

app/helpers/layout_helper.rb
```diff
@@ -4,9 +4,15 @@ def active_org(org)
   end
 
   def colorize_transaction(transaction)
-    css_class = colorize_amount(transaction.amount)
-    if params[:controller] == 'home' && current_member
-      if current_member.last_visited_at
+    if transaction.category && transaction.category.name == 'Receipt'
```
shir: @yunixon just add to transaction model method `transfer?` and use it everywhere. In this method compare category by id. SO it will be something like this:
```rb
def transfer?
  category_id == Category.receipt_id
end
```
in Category
```
def self.receipt_id
  @receipt_id ||= find_by(name: CATEGORY_TRANSFER_...)
end
```

app/views/transactions/create_transfer.js.erb
shir: @yunixon 
```scss
.transfer {
   background-color: $transfer-color;
   &.new-transaction {
     background-color: ...
   }
}
```

app/models/transaction.rb
```diff
@@ -49,6 +51,9 @@ def total
       where('bank_account_transactions.currency' => currency) }
   scope :incomes,     -> { joins(:category).where('categories.type' => Category::CATEGORY_INCOME)}
   scope :expenses,    -> { joins(:category).where('categories.type' => Category::CATEGORY_EXPENSE)}
+  scope :without_out, -> { joins('LEFT OUTER JOIN categories category_transactions
+      ON category_transactions.id = transactions.category_id').
+      where('category_transactions.id IS NULL OR category_transactions.name != ?', Category::CATEGORY_TRANSFER_OUTCOME) }
```
shir: @yunixon instead join you may only load id of transfer category and filter by id
```rb
where(category_id: Category.transfer_outcome_id)
```

app/models/transaction.rb
```diff
@@ -112,6 +117,10 @@ def find_customer_name_by_id(customer_id)
     ''
   end
 
+  def transfer?
+    category_id == Category.receipt_id && category_id.present?
```
shir: @yunixon You may skip condition `category_id.present?` because if `category_id` is nil it will not equal `Category.receipt_id`

yunixon: In tests Category.receipt_id == nil because haven't system categories

PR #241, opened 2015-08-24
Invitations to members

yunixon:


app/views/members/index.html.slim
```diff
@@ -13,5 +11,25 @@ table.table.table-hover.categories
 
 #role_modal_container
 
+- if policy(:invitations).index?
```
shir: @yunixon put invitations stuff in separate partial

PR #245, opened 2015-08-25
transfer out with filter

yunixon:


app/controllers/home_controller.rb
```diff
@@ -6,7 +6,8 @@ def show
     @q = current_organization.transactions.ransack(params[:q])
     @transactions = @q.result
     @curr_flow = @transactions.flow_ordered(current_organization.default_currency) if params[:q]
-    @transactions = @transactions.without_out.page(params[:page]).per(50)
+    @transactions = params[:q] ? @transactions.page(params[:page]).per(50)
```
shir: @yunixon it will work wrong

PR #249, opened 2015-09-02
nested transfer-out

yunixon:


app/models/transaction.rb
```diff
@@ -62,10 +63,12 @@ def total
   validates :customer_name, :comment, length: { maximum: 255 }
   validates :transaction_type, inclusion: { in: TRANSACTION_TYPES, allow_blank: true }
   validates :date, presence: true
+  validate :check_bank_accounts, on: :update, if: Proc.new{ transfer? }
```
shir: @yunixon `if: :transfer?`

app/models/transaction.rb
```diff
@@ -62,10 +63,12 @@ def total
...
 
   before_validation :find_customer, if: Proc.new{ customer_name.present? && bank_account.present? }
   before_validation :set_date, if: Proc.new{ date.blank? }
   before_save :check_negative
+  before_save :sync_date, if: Proc.new{ transfer? }
```
shir: @yunixon same here

PR #252, opened 2015-09-07
fix footer

yunixon:


app/assets/stylesheets/common.css.scss
```diff
@@ -1,7 +1,3 @@
-body {
-  padding-bottom: 130px;
-}
```
AlexanderUlitin: @yunixon keep it in styles as default value.

app/assets/javascripts/transactions.js
```diff
@@ -8,6 +8,12 @@ $(function () {
 
   showHidePeriodAdditionalInput();
 
+  if ($('#flow').length) {
+    $('body').css('padding-bottom', '130px');
+  } else {
+    $('body').css('padding-bottom', '45px');
```
AlexanderUlitin: @yunixon it will be better ti add some class to body tag on that pages, instead of adding this hack in js.

app/assets/stylesheets/common.css.scss
```diff
@@ -1,5 +1,8 @@
 body {
-  padding-bottom: 130px;
+  padding-bottom: 45px;
+  .with_flow {
```
AlexanderUlitin: @yunixon have you tested it? I think it should be `&.with_flow {`.

AlexanderUlitin: Also we use `dash-styled` names for classes, check code conventions.

app/views/home/show.html.slim
```diff
@@ -7,7 +7,11 @@
...
-      .container
-        = render 'home/flow'
+  .footer.navbar.navbar-default.navbar-fixed-bottom#footer
+    .container
+      - if params[:q]
+        .col-sm-9
+          = render 'home/flow'
+      - else
+        .col-sm-12
+          center style=('margin: 20px') Cashbox &copy; #{Date.today.year}
```
AlexanderUlitin: @yunixon move this style to CSS. Don't use inline styles, check our code conventions.

PR #258, opened 2015-09-10
invoice calculate total amount

yunixon:


app/assets/javascripts/invoices.js
```diff
@@ -5,6 +5,12 @@ $(function () {
     .on('cocoon:after-insert', function(e) {
       $('.nested-fields td').css('padding', '8px');
       addCustomerSelect2('.nested-customer-select2:last');
+      $('#invoice_amount').prop('disabled', true);
+    })
+    .on('cocoon:after-remove', function(e) {
+      if ($('#invoice tr.nested-fields:visible').length <= 0) {
+        $('#invoice_amount').prop('disabled', false);
```
AlexanderUlitin: @yunixon let's do it simpler:

```
$('#invoice_amount').prop('disabled', ($('#invoice tr.nested-fields:visible').size() > 0));
```

app/views/invoices/_form.html.slim
```diff
@@ -10,7 +10,7 @@
       input_html: { class: 'datepicker' }
     = f.input :ends_at, as: :string, placeholder: 'End date',
       input_html: { class: 'datepicker' }
-    = f.input :amount
+    = f.input :amount, disabled: @invoice.invoice_items.any? ? true : false
```
AlexanderUlitin: @yunixon you don't need to use ternary operator if your condition returns Boolean. Just use `= f.input :amount, disabled: @invoice.invoice_items.any?`

spec/features/invoices/create_spec.rb
```diff
@@ -55,7 +54,7 @@
       click_on 'Show'
     end
 
-    it { expect(page).to have_css('td', text: money_with_symbol(amount)) }
+    it { expect(page).to have_css('td', text: money_with_symbol(total_amount)) }
```
AlexanderUlitin: @yunixon add checking that amount field is disabled.

spec/features/invoices/update_spec.rb
```diff
@@ -40,16 +40,12 @@
...
-      page.execute_script("$(\"##{dom_id(invoice_with_items, :edit)} #invoice_amount\").val('');")
-      fill_in 'invoice[amount]', with: new_amount
       first('#invoice .nested-fields input.nested-amount').set(new_item_amount)
       first('#invoice .nested-fields textarea.nested-description').set('First Nested Description')
       click_on 'Update Invoice'
     end
 
     it { expect(page).to have_content 'Invoice was successfully updated' }
-    it { expect(page).to have_css("##{dom_id(invoice_with_items)} td",
-      text: money_with_symbol(Money.new(new_amount, invoice.currency))) }
```
AlexanderUlitin: @yunixon add checking that amount is disabled right after opening Edit page, and then after adding/removing InvoiceItems.

PR #261, opened 2015-09-10
invoice - check overlap dates

yunixon:


app/models/invoice.rb
```diff
@@ -32,18 +32,32 @@ class Invoice < ActiveRecord::Base
   validates :ends_at, presence: true
   validates :amount, presence: true
   validates :currency, presence: true
+  validates :customer_name, presence: true
 
   validates :amount, numericality: { greater_than: 0,
     less_than_or_equal_to: Dictionaries.money_max }
   validates :currency, inclusion: { in: Dictionaries.currencies,
     message: "%{value} is not a valid currency" }
+  validate :check_overlap_dates_for_customer, if: 'customer_name.present?'
```
AlexanderUlitin: @yunixon please check this gem https://github.com/robinbortlik/validates_overlap I think it will be better to use this, instead of custom solution.

yunixon: @AlexanderUlitin this gem needs ещ starts_at field present
https://github.com/robinbortlik/validates_overlap/issues/23

app/models/invoice.rb
```diff
@@ -32,13 +32,16 @@ class Invoice < ActiveRecord::Base
...
 
   validates :amount, numericality: { greater_than: 0,
     less_than_or_equal_to: Dictionaries.money_max }
   validates :currency, inclusion: { in: Dictionaries.currencies,
     message: "%{value} is not a valid currency" }
+  validates :starts_at, :ends_at, overlap: { scope: 'customer_id' }
 
-  scope :ordered, -> { order('created_at DESC') }
+  scope :ordered,    -> { order('created_at DESC') }
+  scope :all_except, -> (invoice) { where.not(id: invoice) }
```
AlexanderUlitin: @yunixon you don't use this scope, let's remove it.

spec/models/invoice_spec.rb
```diff
@@ -28,8 +28,47 @@
...
+      let(:invoice3) { build :invoice, customer_name: customer.name,
+        organization: org, starts_at: Date.today - 9.days, ends_at: Date.today }
+      let(:invoice4) { build :invoice, customer_name: customer.name,
+        organization: org, starts_at: Date.today - 11.days, ends_at: Date.today + 1.days }
+      let(:invoice5) { build :invoice, customer_name: customer.name,
+        organization: org, starts_at: Date.today + 2.days, ends_at: Date.today + 10.days }
+
+      it 'Show error on starts_at' do
+        invoice1.valid?
+        expect(invoice1.errors[:starts_at]).to include('overlaps with another record')
```
AlexanderUlitin: @yunixon please check if we can change this message? I would prefer to see error like `overlaps with another Invoice`, if possible.

PR #262, opened 2015-09-14
Add chart by customers

yunixon:


app/models/invoice.rb
```diff
@@ -47,6 +47,25 @@ class Invoice < ActiveRecord::Base
 
   private
 
+  def self.period(period)
+    case period
+    when 'current-month'
+      where('invoices.ends_at >= ? AND invoices.ends_at <= ?', Time.now.beginning_of_month, Time.now)
```
AlexanderUlitin: @yunixon why not use `invoices.ends_at between ? AND ?'` for all cases? I think it's more compact and readable.

AlexanderUlitin: @yunixon please set end of period as `Date.today.end_of_month`.

AlexanderUlitin: @yunixon another thing: if you set this fields as `date` in database schema, why you use time in conditions? Maybe it will be better to use Date too? like `where('invoices.ends_at between ? AND ?', Date.current.beginning_of_month, Date.current.end_of_month)`

app/models/invoice.rb
```diff
@@ -47,6 +47,25 @@ class Invoice < ActiveRecord::Base
...
+    when 'last-3-months'
+      where('invoices.ends_at >= ? AND invoices.ends_at <= ?', (Time.now - 3.months).beginning_of_day, Time.now)
+    when 'prev-month'
+      prev_month_begins = Time.now.beginning_of_month - 1.months
+      where('invoices.ends_at between ? AND ?', prev_month_begins,
+        prev_month_begins.end_of_month)
+    when 'this-year'
+      where('invoices.ends_at >= ? AND invoices.ends_at <= ?', Time.now.beginning_of_year, Time.now)
+    when 'quarter'
+      where('invoices.ends_at >= ? AND invoices.ends_at <= ?', Time.now.beginning_of_quarter, Time.now)
```
AlexanderUlitin: @yunixon please also set end of period as accurate date, not current date or `Time.now`.

app/models/invoice.rb
yunixon: @AlexanderUlitin I take this class method from transaction.rb :)

app/models/invoice.rb
```diff
@@ -47,6 +47,25 @@ class Invoice < ActiveRecord::Base
 
   private
 
+  def self.period(period)
+    case period
+    when 'current-month'
+      where('invoices.ends_at between ? AND ?', Date.today.beginning_of_month, Date.today.end_of_month)
```
AlexanderUlitin: @yunixon it's better t use `Date.current`. 
http://makandracards.com/makandra/4601-use-time-current-date-current-datetime-current
http://stackoverflow.com/questions/6635363/what-is-the-difference-between-date-current-and-date-today

Please fix it in all places.

app/helpers/layout_helper.rb
```diff
@@ -19,6 +19,12 @@ def show_amount_with_tooltip(amount)
     end
   end
 
+  def show_tooltip_with_text(text)
```
AlexanderUlitin: @yunixon please refactor method `show_amount_with_tooltip`, so there will be no duplicated code.

app/models/invoice.rb
yunixon: @AlexanderUlitin In all places of project or this method only?

AlexanderUlitin: @yunixon in all project, it's not too many places.

PR #274, opened 2015-09-16
show invoice as pdf

yunixon:


app/views/layouts/pdf.html.slim
```diff
@@ -0,0 +1,9 @@
+doctype html
+html
+  head
+    title Cashbox
+    = wicked_pdf_stylesheet_link_tag    'application', media: 'all'
+    = wicked_pdf_javascript_include_tag 'application'
```
AlexanderUlitin: @yunixon do we need js here?

app/views/layouts/pdf.html.slim
```diff
@@ -0,0 +1,9 @@
+doctype html
+html
+  head
+    title Cashbox
+    = wicked_pdf_stylesheet_link_tag    'application', media: 'all'
```
AlexanderUlitin: @yunixon I think we need to add separated `pdf.css` file with stylesheets only for PDF.

PR #275, opened 2015-09-17
create transaction by invoice

yunixon:


app/controllers/transactions_controller.rb
```diff
@@ -5,14 +5,21 @@ class TransactionsController < ApplicationController
...
+      @transfer = Transfer.new
+    end
   end
 
   def create
     @transaction = Transaction.new(transaction_params)
     check_relation_to_curr_org(:transaction)
-    @transaction.save
+    if @transaction.save
+      @transaction.invoice.update(paid_at: @transaction.date) if @transaction.invoice
```
AlexanderUlitin: @yunixon please move this logic to model.

app/views/transactions/_form.html.slim
```diff
@@ -29,14 +29,21 @@
         collection: current_organization.bank_accounts.visible.positioned.grouped_by_currency(current_organization.default_currency),
         prompt: 'Bank account', group_method: :last, label: 'From',
         selected: t.object.bank_account_id
-
+  - elsif params[:invoice_id] && params[:customer_id] || f.object.invoice_id
+    = f.input :customer_name, as: :hidden
```
AlexanderUlitin: @yunixon I think it will be better to not hide this field, but make is disabled, so we will be informed what customer will be set to transaction.

app/views/transactions/_form.html.slim
```diff
@@ -46,6 +53,16 @@
   - else
     = f.input :bank_account, disabled: true
 
-  = f.input :comment
+  - if params[:invoice_id] || f.object.invoice_id
+    = f.input :comission, input_html: { class: 'comission' }
+
+  - if params[:invoice_id] || f.object.invoice_id
```
AlexanderUlitin: @yunixon let's keep this field as editable and not hidden.

app/models/transaction.rb
```diff
@@ -148,15 +155,30 @@ def set_date
...
+      humanized_money_with_symbol(Money.new(comission.to_d * 100, bank_account.currency),
+      symbol_after_without_space: true)
+  end
+
   def residue?
     self.transaction_type == 'Residue'
   end
 
+  def update_invoice_paid_at
+    self.invoice.update(paid_at: self.date)
```
AlexanderUlitin: @yunixon need to make sure that we will not update Invoice in other organization. Please add conditions to this query, so we will search Invoice only in Transaction Organization.

app/controllers/transactions_controller.rb
```diff
@@ -5,8 +5,13 @@ class TransactionsController < ApplicationController
 
   def new
     @q = current_organization.transactions.ransack(session[:filter])
-    @transaction = Transaction.new
-    @transfer = Transfer.new
+    if params[:invoice_id]
```
AlexanderUlitin: @yunixon please make sure here that invoice_id is correct and it's a real Invoice from current_organization. User `before_filter` with set `@invoice` variable. Then use this preparing of @transaction fields only if `@invoice.present?`

app/controllers/transactions_controller.rb
```diff
@@ -5,8 +5,13 @@ class TransactionsController < ApplicationController
...
+      @transaction = Transaction.new(customer_name: current_organization.find_customer_name_by_id(params[:customer_id]),
+        amount: params[:amount], invoice_id: params[:invoice_id])
```
AlexanderUlitin: @yunixon use `amount` and `customer` from `@invoice` object. Don't trust params from users.

app/views/transactions/_form.html.slim
```diff
@@ -29,14 +29,21 @@
...
+  - elsif params[:invoice_id] && params[:customer_id] || f.object.invoice_id
+    = f.input :customer_name, disabled: true
   - else
     = f.input :customer_name, input_html: { class: 'customer-select2', data: { url: autocomplete_customers_path,
       placeholder: 'Customer',
       value: @q.nil? ? current_organization.find_customer_name_by_id(f.object.customer_id) \
         : (@q.customer_id_eq ? current_organization.find_customer_name_by_id(@q.customer_id_eq) : nil) } }
 
-  - if f.object.bank_account.nil? || f.object.bank_account.visible?
+  - if params[:invoice_id] && params[:currency] || f.object.invoice_id
```
AlexanderUlitin: @yunixon use `currency` from `@invoice`. Don't trust params from users.

app/views/transactions/_form.html.slim
```diff
@@ -29,14 +29,21 @@
...
+    = f.association :bank_account, as: :grouped_select,
+      collection: current_organization.bank_accounts.where(currency: params[:currency] || \
```
AlexanderUlitin: @yunixon same here.

app/views/transactions/_form.html.slim
```diff
@@ -46,6 +53,13 @@
   - else
     = f.input :bank_account, disabled: true
 
+  - if params[:invoice_id] || f.object.invoice_id
```
AlexanderUlitin: @yunixon don't use params here, use `@invoice` or even better `f.object.invoice_id`

app/views/transactions/_form.html.slim
```diff
@@ -46,6 +53,13 @@
...
+    = f.input :comission, input_html: { class: 'comission' }
+
   = f.input :comment
+
   = f.input :date, as: :string, placeholder: 'Date',
     input_html: { class: 'datepicker', value: f.object.date.present? ? l(f.object.date.to_date) : nil }
+
+  - if params[:invoice_id] || f.object.invoice_id
```
AlexanderUlitin: @yunixon same here, use `@invoice` or even better `= f.hidden_field :invoice`

spec/features/invoices/index_spec.rb
```diff
@@ -35,4 +35,42 @@
...
+  context 'complete invoice', js: true do
+    let!(:account)  { create :bank_account, organization: org }
+    let!(:category) { create :category, :income, organization: org }
+    let!(:invoice)  { create :invoice, organization: org, amount: 500 }
+    let(:comission) { Money.new(100, invoice.currency) }
+
+    before do
+      visit invoices_path
+      click_on 'Complete Invoice'
+      within '#new_transaction' do
```
AlexanderUlitin: @yunixon add extra case on this page, with checking form. Make sure in spec that form shows only Income categories, only Accounts with Invoice.currency. Show Amount and Commission fields.

spec/features/invoices/index_spec.rb
```diff
@@ -35,4 +35,42 @@
...
+        select category.name, from: 'transaction[category_id]'
+        select account.name, from: 'transaction[bank_account_id]'
+        fill_in 'transaction[comission]', with: comission
+        fill_in 'transaction[comment]', with: 'TestComment'
+      end
+      click_on 'Create'
```
AlexanderUlitin: @yunixon right after clicking on `Create` button - you need to add check for new added Transaction on the page with proper amount. You can also check Transaction record in database, but check it right after creating is important case.

spec/features/invoices/index_spec.rb
```diff
@@ -35,4 +35,42 @@
...
+      visit invoices_path
```
AlexanderUlitin: @yunixon move `visit invoices_path` for separated case. You should only Create Transaction in this before block.

app/views/transactions/_form.html.slim
```diff
@@ -46,6 +53,13 @@
   - else
     = f.input :bank_account, disabled: true
 
+  - if params[:invoice_id] || f.object.invoice_id
+    = f.input :comission, input_html: { class: 'comission' }
```
AlexanderUlitin: @yunixon please add `Amount` in separated line on this form (not like text input, just like a text, but in correct currency and format), so user will see Amount of new transaction, that he is going to create on this form.

app/views/transactions/_form.html.slim
```diff
@@ -6,7 +6,7 @@
 
   - if f.object.category.try(:system?)
     = f.input :category, disabled: true
-  - elsif params[:action] == 'new' || params[:action] == 'create'
+  - elsif params[:action] == ('new' || 'create')
```
AlexanderUlitin: @yunixon it's a wrong code. If you want to refactor it - you can use `['new', 'create'].include?(params[:action])`. Please check specs, cos I think they are wrong too, if there are no failed specs with this code.

app/views/transactions/_form.html.slim
```diff
@@ -18,7 +18,7 @@
       group_method: :last, prompt: 'Category', value_method: :id,
       selected: f.object.category_id
 
-  - if transaction.transfer? && (params[:action] == 'edit' || params[:action] == 'update')
+  - if transaction.transfer? && params[:action] == ('edit' || 'update')
```
AlexanderUlitin: @yunixon it's a wrong code, please fix it.

app/views/transactions/_form.html.slim
```diff
@@ -46,6 +54,14 @@
   - else
     = f.input :bank_account, disabled: true
 
+  - if (@invoice.present? || f.object.invoice_id) && params[:action] == ('new' || 'create')
```
AlexanderUlitin: @yunixon it's a wrong code, please fix and check related specs.

spec/features/invoices/index_spec.rb
```diff
@@ -46,13 +44,73 @@
...
+      end
+      expect(page).to_not have_select('Category', with_options: [wrong_category.name])
+      expect(page).to_not have_select('Bank Account', with_options: [wrong_account.name])
+      expect(page).to have_field('Comission', with: comission)
+      expect(page).to have_content("Total amount: #{invoice.amount - comission}" )
+      click_on 'Create'
+      sleep 1 # wait after page rerender
+    end
+
+    subject{ create_transaction_by_invoice; page }
```
AlexanderUlitin: @yunixon looks like you use this subject only in context 'update invoice paid_at after create transaction'. Please define `subject` inside that context only.

spec/features/invoices/index_spec.rb
```diff
@@ -46,13 +44,73 @@
...
+        select account.name, from: 'transaction[bank_account_id]'
+        fill_in 'transaction[comission]', with: comission
+        fill_in 'transaction[comment]', with: 'TestComment'
+      end
+      expect(page).to_not have_select('Category', with_options: [wrong_category.name])
+      expect(page).to_not have_select('Bank Account', with_options: [wrong_account.name])
+      expect(page).to have_field('Comission', with: comission)
+      expect(page).to have_content("Total amount: #{invoice.amount - comission}" )
+      click_on 'Create'
+      sleep 1 # wait after page rerender
```
AlexanderUlitin: @yunixon you can avoid of adding `sleep`. Just use that methods like `click_link` - it has timeouts for waiting while page load (check source code of Capybara helpers if need).

AlexanderUlitin: Anyway it's a bad practice to use `sleep`.

spec/features/invoices/index_spec.rb
```diff
@@ -46,13 +44,73 @@
...
+    subject{ create_transaction_by_invoice; page }
+
+    context 'update invoice paid_at after create transaction' do
+      before do
+        create_transaction_by_invoice
+        visit invoices_path
+      end
+
+      it 'invoice paid_at must present' do
+        expect(page).to have_content(invoice.paid_at)
```
AlexanderUlitin: @yunixon please check exact row and exact column in the table with this content. There might be another date on the date, that just match with paid_at.

spec/features/invoices/index_spec.rb
```diff
@@ -46,13 +44,73 @@
...
+    def create_transaction_by_invoice
+      visit invoices_path
+      click_on 'Complete Invoice'
+      within '#new_transaction' do
+        select category.name, from: 'transaction[category_id]'
+        select account.name, from: 'transaction[bank_account_id]'
+        fill_in 'transaction[comission]', with: comission
+        fill_in 'transaction[comment]', with: 'TestComment'
+      end
+      expect(page).to_not have_select('Category', with_options: [wrong_category.name])
```
AlexanderUlitin: @yunixon move all this expects out of this function. You don't need to repeat this in all places.

spec/features/invoices/index_spec.rb
yunixon: @AlexanderUlitin not only this context, its used also in 'with valid data' context

app/models/transaction.rb
```diff
@@ -148,15 +155,30 @@ def set_date
     self.date = Time.current
   end
 
+  def calculate_amount
+    self.amount = Money.new(amount_cents - comission.to_d * 100, currency)
+    add_comission_to_comment
```
AlexanderUlitin: @yunixon I think it will be better to call this method with `!` symbol, cos it makes updates in the class, so it should looks like: `self.add_comission_to_comment!`

AlexanderUlitin: Sorry, nevermind, I was wrong, cos there is no saving.

PR #276, opened 2015-09-17
Fix localize date bug

yunixon:


AlexanderUlitin:
@yunixon is there any way we can cover this bug in specs?

PR #281, opened 2015-09-23
Remove transfers from flow sum table

yunixon:


AlexanderUlitin:
@yunixon please cover this case in specs.

spec/features/transactions_filter_spec.rb
```diff
@@ -404,5 +405,20 @@
...
+      let!(:transfer)    { create :transfer, :with_different_currencies, bank_account_id: ba.id }
+
+      before do
+        visit root_path
+        within "##{dom_id(transaction)}" do
+          click_link ba.to_s
+        end
+      end
+
+      it { expect(page).to have_content("Total: #{amount}") }
```
AlexanderUlitin: @yunixon you should check this content not in whole page. You need to check that it showed in specific cell of totals table.

PR #287, opened 2015-10-02
Slide toggle filter

yunixon:


app/assets/javascripts/transactions.js
```diff
@@ -73,6 +74,10 @@ $(function () {
       $('#edit_transaction_' + trans_id).submit();
     }
   });
+
+  $(document).on('click', '#toggle_filter', function(e) {
+    $('#transaction_search').slideToggle();
```
AlexanderUlitin: @yunixon add `e.preventDefault();` to avoid of adding `#` to the page url.

app/assets/javascripts/transactions.js
```diff
@@ -1,4 +1,5 @@
 $(function () {
+  $('#transaction_search').hide();
```
AlexanderUlitin: @yunixon we need to hide filters only if we have not used it. If we filtered transactions - we need to keep Filters showed.

PR #299, opened 2015-10-06
Fix create transfer form: save calculate sum

yunixon:


app/views/transactions/_transfer_form.html.slim
```diff
@@ -12,7 +12,9 @@
     .form-group.string.optional.transfer_sum
       label for='end_sum' class='string optional control-label col-sm-3' Calculate sum
       .col-sm-9
-        input type='text' class='string optional form-control' id='end_sum' placeholder='Sum'
+        input type='text' class='string optional form-control' value="#{Money.new(\
```
AlexanderUlitin: @yunixon I think it's a wrong way to keep value in this field in case errors. Why not add virtual attribute and use it? Also please cover this case in specs.

PR #309, opened 2015-10-08
Fix removing EUR from currencies Dictionary

yunixon:


spec/models/organization_spec.rb
```diff
@@ -33,10 +33,6 @@
       Dictionaries.currencies << 'EUR' unless Dictionaries.currencies.include?('EUR')
     end
 
-    after do
-      Dictionaries.currencies.delete('EUR') if Dictionaries.currencies.include?('EUR')
```
AlexanderUlitin: @yunixon I think it's wrong fix. In this spec we need to use some currency that we don't have implemented in rates. Use AED (Emirati Dirham) for example.

spec/models/organization_spec.rb
```diff
@@ -30,24 +30,26 @@
...
+      Dictionaries.currencies << 'JPY' unless Dictionaries.currencies.include?('JPY')
     end
 
     after do
-      Dictionaries.currencies.delete('EUR') if Dictionaries.currencies.include?('EUR')
+      Dictionaries.currencies.delete('JPY') if Dictionaries.currencies.include?('JPY')
     end
 
-    context 'with updated Dictionary for including EUR' do
+    context 'with updated Dictionary for including AED' do
```
AlexanderUlitin: @yunixon fix context name.

PR #312, opened 2015-10-08
Fix statistic charts

yunixon:


app/assets/javascripts/statistics.js
```diff
@@ -141,6 +141,45 @@ var drawBalanceChart = function drawBalanceChart(period, element) {
   }
 };
 
+var drawCustomersBalanceChart = function drawCustomersBalanceChart(period, element) {
```
AlexanderUlitin: @yunixon I see 2 almost similar functions here. Let's combine it into one function, but with some params (for add some extra logic in case you need this).

app/assets/javascripts/statistics.js
```diff
@@ -119,11 +119,17 @@ var drawBalanceChart = function drawBalanceChart(period, element) {
...
+    if (css_id.id == 'main-balance') {
+      chart = new google.visualization.ComboChart(css_id);
+    } else {
+      chart = new google.visualization.ColumnChart(css_id);
+    }
     var formatter = new google.visualization.NumberFormat(response.currency_format);
     formatter.format(chartData, 1);
     formatter.format(chartData, 2);
-    formatter.format(chartData, 3);
+    if (css_id.id == 'main-balance') { formatter.format(chartData, 3) };
```
AlexanderUlitin: @yunixon we don't use conditions in one line in JS. Please check our code-conventions.

PR #315, opened 2015-10-12
Fix members layout: pull right buttons

yunixon:


app/assets/stylesheets/common.css.scss
```diff
@@ -73,3 +73,7 @@ body {
 .invoice-title {
   color: $title-invoice-color
 }
+
+#new_invitation {
```
AlexanderUlitin: @yunixon it's a usual ID for form. Please use something more specific for the link, like `#invitation_link` or `#invitation_link_area`.

AlexanderUlitin: I see that form for create new Invite will have this margin too, I'm not sure if you need this.

PR #331, opened 2016-02-02
Refactoring invoices view

yunixon:


app/assets/javascripts/invoices.js
```diff
@@ -17,6 +17,10 @@ $(function () {
   $(document).on('change', '.nested-amount', function(e) {
     calculateInvoiceAmount();
   });
+
+  $(document).on('click', '.invoice[data-href]', function(e) {
```
AlexanderUlitin: @yunixon have you seen event for `[data-edit-url]`? Maybe you can use it, instead of duplicating functionality?

yunixon: @AlexanderUlitin I think its different functional, in function with [data-edit url] we use ajax with dataType: "script" for open modal popup, in this function we need to redirect to show invoice page.

AlexanderUlitin: @yunixon ok, got it, thanks.

PR #358, opened 2016-08-12
Add all customers and totlal summs to invoices index page

dissident:
https://trello.com/c/MIoyFsIB/341--

### DIFF
--- a/app/helpers/layout_helper.rb
+++ b/app/helpers/layout_helper.rb
@@ -28,7 +28,7 @@ module LayoutHelper
 
   def invoices_debt(debtor)
     if debtor.is_a? Customer
-      str = "#{debtor.name}:"
+      str = h "#{debtor.name}:"
     else
       str = "All customers:"
     end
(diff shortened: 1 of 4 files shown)


app/helpers/layout_helper.rb
```diff
@@ -27,7 +27,11 @@ def submit_title
   end
 
   def invoices_debt(debtor)
-    str = "#{debtor.name}:"
+    if debtor.is_a? Customer
+      str = "#{debtor.name}:"
```
shir: @dissident add html escaping. This string is defined as html safe on return, but customer name is not html safe.

app/helpers/layout_helper.rb
```diff
@@ -45,4 +49,22 @@ def invoices_debt(debtor)
 
     str.html_safe
   end
+
+  def total_invoices_debt
+    str = "<strong>Total: "
+    total = 0
+    organization = current_organization
+    def_curr = organization.default_currency
+    organization.invoices.unpaid.group(:currency).sum(:amount_cents).each do |currency, amount_cents|
```
shir: @dissident move this calculation to model

PR #366, opened 2016-08-31
Stas/action cable notifications

dissident:
https://trello.com/c/3wn66dwg/57--


config/cable.yml
```diff
@@ -1,8 +1,10 @@
 development:
```
shir: @dissident remove this file from git, add it to gitignore and create sample file instead.

PR #368, opened 2016-09-06
Add copy button on transaction edit modal window

dissident:
https://trello.com/c/8w9iB8et/212--


app/controllers/transactions_controller.rb
```diff
@@ -42,6 +42,25 @@ def destroy
     @transaction.destroy
   end
 
+  def copy
```
shir: @dissident instead new action use same `create` action but use some query parameter like `copy_transaction_id` and if this parameter exists then prefill builded transaction with data from that transaction.

PR #369, opened 2016-09-07
Stas/noclosed transaction form

dissident:
https://trello.com/c/6L998Ssd/346--


app/controllers/transactions_controller.rb
```diff
@@ -42,6 +44,25 @@ def destroy
     @transaction.destroy
   end
 
+  def copy
+    if @transaction.transfer?
+      @transaction = @transaction.dup
```
AlexanderUlitin: @dissident I'm not sure if it's correct way... it will probably copy `created_at` and `updated_at` too? I think we need to copy only specific fields, like Category, Customer, Amount, Bank Account, Comment, Date.

AlexanderUlitin: @shir what do you think about it?

dissident: @AlexanderUlitin I am not sure, but I think this fields will be overwritten after save.

dissident: @AlexanderUlitin, I checked, the fields `:created_at` and `:updated_at` are overwritten.

PR #377, opened 2016-09-16
Change filter link

yunixon:


spec/features/transactions_filter_spec.rb
```diff
@@ -27,7 +27,7 @@
 
     before do
       visit root_path
-      click_on 'Filters'
+      find(:css, 'a#toggle_filter').trigger('click')
```
shir: @yunixon you may still reference link by title. So `click_on 'Filters'` should work.

yunixon: @shir no, with js: true gets error: Failed to click element

shir: @yunixon it should work with `js: true` too. I used it many times in other projects

yunixon: @shir Capybara::Webkit::ClickFailed: Failed to click element /html/body/div/div[2]/div[1]/h2/a[@id='toggle_filter'] because of overlapping element /html/body/div/div[2]/div[1]/h2

yunixon: https://github.com/thoughtbot/capybara-webkit/issues/494#issuecomment-221616682

PR #385, opened 2016-09-17
Add disable_with

yunixon:


config/initializers/simple_form_disable_with.rb
```diff
@@ -0,0 +1,14 @@
+module DisableDoubleClickOnSimpleForms
```
shir: @yunixon no. Remove this code. It should not be done this way. When other programmer he will not expect that diable_with will be applied. If you want make less repeating, then simple form allow to define new types for buttons, so you may define something like `submit_with_disable` and use it `f.button :submit_with_disable` everywhere. For example https://gist.github.com/maxivak/57a2fb71afeb9efcf771

PR #387, opened 2016-09-19
Add base api and jwt auth

yunixon:


app/controllers/api/base_controller.rb
```diff
@@ -0,0 +1,29 @@
+module Api
```
shir: @yunixon move it to `/api/v1/`

Gemfile
```diff
@@ -51,6 +51,8 @@ gem 'wkhtmltopdf-binary'
 gem 'puma', '~> 3.0'
 gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
 gem 'redis'
+gem 'knock', '~> 1.4.2'
```
shir: @yunixon As I remember latest version is 1.5. Why do you use 1.4.2?

app/controllers/api/v1/auth_token_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+          return
+        else
+          render json: { error: I18n.t("devise.failure.locked") }, status: :unauthorized
+        end
+      else
+        raise Api::BaseController::AuthorizationError
+      end
+    end
+
+    def auth_token
```
shir: @yunixon not sure why do you need all this code. Knock gem hold all of this.

app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,9 @@
+module Api::V1
+  class UsersController < ApiController
+
+    def show
```
shir: @yunixon move it to action `current`. I'm pretty sure we will need `show` action in future to get user by id.

lib/api_constraints.rb
```diff
@@ -0,0 +1,10 @@
+class ApiConstraints
+  def initialize(options)
+    @version = options[:version]
+    @default = options[:default]
+  end
+
+  def matches?(req)
+    @default || req.headers['Accept'].include?("application/vnd.sagn.v#{@version}")
```
shir: @yunixon why do you need this?

config/initializers/active_model_serializer.rb
```diff
@@ -0,0 +1 @@
+ActiveModel::Serializer.config.adapter = :json
```
shir: @yunixon why do you need this?

app/serializers/user_serializer.rb
```diff
@@ -0,0 +1,26 @@
...
+#  last_sign_in_ip        :string(255)
+#  failed_attempts        :integer          default(0), not null
+#  unlock_token           :string(255)
+#  locked_at              :datetime
+#  created_at             :datetime
+#  updated_at             :datetime
+#  full_name              :string(255)      not null
+#
+
+class UserSerializer < ActiveModel::Serializer
```
shir: @yunixon instead it's better to use `jbuilder` gem.

Gemfile
yunixon: 1.5 version has issue with rails 5: https://github.com/nsarno/knock/issues/104

shir: Hmm... Strange. I use in on 5.0 in other project

shir: Then add link to issue in comment in Gemfile.

app/controllers/api/base_controller.rb
yunixon: @shir this is general base controller for different versions, I think it should be in api folder

shir: @yunixon authorization methods (and this controller holds only authorization method) should be in v1 because they may change in future versions. And same with errors handling. For example in next version we may decide to use xml. Or SOAP or something else.

app/serializers/user_serializer.rb
yunixon: @shir Why? I think better to use serializers
http://kirillplatonov.com/2014/11/04/active_model_serializer_vs_jbuilder/

shir: @yunixon because jbuilder is much more flexible. I tried other solutions in different projects and found that at some point you meet some limits.

app/controllers/api/v1/auth_token_controller.rb
yunixon: Knock token controller (Knock::AuthTokenController) not working with 1.4.2

app/views/api/v1/users/_user.json.jbuilder
```diff
@@ -0,0 +1,5 @@
+json.user do
```
shir: @yunixon you may write instead `json.(user, :id, :email, :full_name)`

app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,9 @@
+module Api::V1
+  class UsersController < ApiController
+
+    def current
+      render partial: 'user', locals: { user: current_user }
```
shir: @yunixon it's better to not make any render here or render `show` action (which we will have in future) and in `current.json.jbuilder` (or `show.json.jbuilder`) make `json.partial! 'user', user: current_user`

lib/api_constraints.rb
yunixon: needs to versioning api by headers, fixed to 'vnd.cashbox.v#'

app/views/api/v1/users/_user.json.jbuilder
yunixon: @shir thats work: json.user(user, :id, :email, :full_name)

shir: @yunixon it works just `json.(...)` too. ;)

yunixon: @shir nope :), i need json: { user: { id:...}} insted of {id: ...}

shir: @yunixon ah... Ok. Didn't notice that.

PR #388, opened 2016-09-19
Fix live_open check_box in transaction and transfer forms

dissident:
https://trello.com/c/TQtSAWvd/361-fix-style-for-leave-open-checkbox


app/views/transactions/_form.html.slim
```diff
@@ -70,5 +70,5 @@
     = link_to "Created from invoice", @invoice, class: 'pull-right'
 
   = f.input :invoice_id, as: :hidden
-  - if transaction.new_record?
-    = f.input :leave_open, as: :boolean
+  - unless transaction.persisted? || @invoice.present? || f.object.invoice_id
+    = f.input :leave_open, as: :boolean, wrapper: :checkbox, wrapper_html: { class: 'col-sm-6 pull-left' }
```
shir: @dissident I guess instead you should use proper wrapper on form. See example https://github.com/rafaelfranca/simple_form-bootstrap/blob/master/app/views/examples/_horizontal_form_sf.html.erb

PR #390, opened 2016-09-19
Fix balance pagination

dissident:
https://trello.com/c/Fxrlz8AM/296-balance

### DIFF
--- a/app/assets/javascripts/statistics.js
+++ b/app/assets/javascripts/statistics.js
@@ -37,12 +37,19 @@ $(function () {
         setStep(step);
         drawBalanceChart(null, scale, step, 'main-balance');
       }
+      toggleStep('.left_step_wrapper', true);
     }
     else if ($(this).hasClass('left-step')){
       step += 1;
       setStep(step);
       drawBalanceChart(null, scale, step, 'main-balance');
     };
+    if (step === 0) {
+      toggleStep('.right_step_wrapper', false);
+    }
+    if (step === 1) {
+      toggleStep('.right_step_wrapper', true);
+    }
   });
 });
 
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/statistics.js
```diff
@@ -64,6 +71,18 @@ var getScale = function getScale() {
   return scale
 }
 
+var prohibitStep = function prohibitStep(elementSelector) {
```
shir: @dissident if you assign function to variable you don't need to name it. So:
```js
var prohibitStep = function (..) { ... }
```
or
```js
function prohibitStep(...) {...}
```

shir: @dissident I think you may write it simpler using `toggleClass`:
```js
function toggleStep(elementSelector, state) {
  element = $(elementSelector);
  element.find('a').toggleClass('hide', state);
  element.find('span').toggleClass('hide', state);
}
```
And then use
```js
toggleStep('.right_step_wrapper', step == 0)
```

dissident: @shir it's not working correctly for my case. One state can be false or true, and I need add class hide to one element and delete from another, but the state in this toggleClass work another way, he adds the only class if true, and remove only if false.

shir: @dissident 
```js
element.find('a').toggleClass('hide', state);
element.find('span').toggleClass('hide', !state);
```

app/assets/javascripts/statistics.js
```diff
@@ -64,6 +71,12 @@ var getScale = function getScale() {
   return scale
 }
 
+function toggleStep(elementSelector, state) {
+  element = $(elementSelector);
```
shir: @dissident didn't notice it before. You miss `var` here too.

PR #393, opened 2016-09-20
Fix balance scale

dissident:
https://trello.com/c/w6fbAgkb/297-balance

### DIFF
--- a/app/assets/javascripts/statistics.js
+++ b/app/assets/javascripts/statistics.js
@@ -155,6 +155,9 @@ var drawBalanceChart = function drawBalanceChart(period, scale, step, element) {
         step: step,
         period: period,
         scale: scale
+      },
+      beforeSend: function (xhr) {
+        toggleBalanceSpinner(false);
       }
     })
     .done(function(response) {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/statistics.js
```diff
@@ -293,3 +306,8 @@ Number.prototype.format = function(n, x) {
   var re = '(\\d)(?=(\\d{' + (x || 3) + '})+' + (n > 0 ? '\\.' : '$') + ')';
   return this.toFixed(Math.max(0, ~~n)).replace(new RegExp(re, 'g'), '$1,');
 };
+
+function toggleBalanceSpinner(state) {
+  spinner = $('.balance-chart.spinner-backround');
```
shir: @dissident
1. Use `var spinner =`
2. I think you may make directly `$('.balance-chart.spinner-backround').toggleClass('hide', state);`

dissident: @shir before coding in mind I plan more code in this function (disable buttons and etc.), but this not useful. I remove `spinner =`.

shir: @dissident main issue was that you forgot `var`

PR #394, opened 2016-09-20
Add api for organizations

yunixon:


app/models/user.rb
```diff
@@ -55,4 +55,16 @@ def authenticate(password)
   def locked?
     self.locked_at.present?
   end
+
+  def current_organization
+    self.organizations.find(self.current_organization_id)
+  rescue
```
shir: @yunixon try not to use `rescue`. It works slow. In this case it's better to check if current_organization_id is blank and if it is return first organization.

app/views/api/v1/organizations/_organization.json.jbuilder
```diff
@@ -0,0 +1 @@
+json.organization organization, :id, :name, :default_currency
```
shir: @yunixon it's better not to include `organization: ...` in json in this partial. In this case you can't use this parial to render collection. If you want then put `organization: ...` in upper level view.

app/views/api/v1/users/_user.json.jbuilder
```diff
@@ -1,5 +1 @@
-json.user do
-  json.id user.id
-  json.email user.email
-  json.full_name user.full_name
-end
+json.user user, :id, :email, :full_name
```
shir: @yunixon same here

db/migrate/20160919080348_add_current_organization_id_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddCurrentOrganizationIdToUsers < ActiveRecord::Migration[5.0]
+  def change
+    add_column :users, :current_organization_id, :integer
```
shir: @yunixon I think we don't need this. Selection organization for which get data is frontend part. Now this store it in cookies. And API client will decide for which organization he want to get data.

PR #403, opened 2016-09-26
Add invites in project without organization creation

dissident:
https://trello.com/c/B9jUM0fB/354--


app/controllers/invitations_to_organization_controller.rb
```diff
@@ -0,0 +1,49 @@
+class InvitationsToOrganizationController < ApplicationController
```
shir: @dissident I think it will be better to name it `OrganiztaionInvitationsContoller`. Same for model

app/controllers/invitations_global_controller.rb
```diff
@@ -0,0 +1,45 @@
+class InvitationsGlobalController < ApplicationController
```
shir: @dissident I think it will be better to name it just `InvitationsController`. Same for model. I know that we already have model with `Invitation` name it may produce some issues. But it will be better to name it right now.

BTW, can we combine both models? Just use empty `organization_id` field for global invitations.

app/models/invitation_to_organization.rb
```diff
@@ -0,0 +1,50 @@
...
+  validate :role_inclusion
+  validate :email_uniq
+
+  delegate :organization, to: :member
+
+  after_create :send_invitation
+
+  enumerize :role, in: [:user, :admin, :owner], default: :user, predicates: true
+
+  def owner
```
shir: @dissident rename to `owner_name`. And I think you may make `owner` as association through `member`

app/models/invitation_to_organization.rb
```diff
@@ -0,0 +1,50 @@
...
+    Member.find(self.invited_by_id).user.to_s
```
shir: @dissident why do you `Member.find(self.invited_by_id)` if there is already `member` association.

app/models/invitation_to_organization.rb
```diff
@@ -0,0 +1,50 @@
...
+  rescue
```
shir: @dissident it's better to check for non null value than catch exception. Exceptions are pretty slow.

app/models/invitation_to_organization.rb
```diff
@@ -0,0 +1,50 @@
...
+    ''
+  end
+
+  def accept!(user)
+    user_member = Member.find_or_create_by!(user: user, organization_id: member.organization_id)
+    user_member.update_attribute(:role, role)
```
shir: @dissident this two strings generate two requests to db. Instead you may do it with one:
```rb
Member.find_or_initialize_by!(user: user, organization_id: member.organization_id).tap do |member|
  member.role = role
  member.save
end
```

app/models/invitation_to_organization.rb
```diff
@@ -0,0 +1,50 @@
...
+    update_attribute(:accepted, true)
+  end
+
+  def send_invitation
+    InvitationMailer.new_invitation_to_organization(self).deliver_now
+  end
+
+  def congratulation
+    "You joined #{member.organization.name}."
```
shir: @dissident it's view part. It should not be in model

app/models/invitation_to_organization.rb
```diff
@@ -0,0 +1,50 @@
...
+  def role_inclusion
+    if invited_by_id
+      unless member.available_roles.to_h.values.include?(role)
+        errors.add(:role, "Should be in #{member.available_roles.to_h.keys.join(', ')}")
+      end
+    end
+  end
+
+  def email_uniq
+    unless organization.invitations.where(email: email).empty?
```
shir: @dissident why can't you check it with uniqueness validation?

app/models/invitation_to_organization.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def congratulation
+    "You joined #{member.organization.name}."
+  end
+
+  private
+
+  def role_inclusion
+    if invited_by_id
+      unless member.available_roles.to_h.values.include?(role)
```
shir: @dissident why can't you check it with inclusion validation?

db/migrate/20160923053347_remove_nill_false_from_invitations.rb
```diff
@@ -0,0 +1,15 @@
+class RemoveNillFalseFromInvitations < ActiveRecord::Migration[5.0]
+  def up
+    change_column :invitations, :invited_by_id, :integer,
```
shir: @dissident why do you change it? I believe that we always have a user how make invitation

app/models/invitation_to_organization.rb
```diff
@@ -0,0 +1,50 @@
+class InvitationToOrganization < Invitation
+  belongs_to :member, inverse_of: :created_invitations, foreign_key: :invited_by_id
```
shir: @dissident I think `invited_by` will be better name for this association.

app/controllers/invitations_global_controller.rb
dissident: @shir I shared these models because they have different validation, different behavior of the same methods, I was trying to avoid a large number of `if` and `case` states. This is my main aim.

app/models/invitation_to_organization.rb
dissident: @shir I write one method in this model, another code I move from old model Invitation. I don't refactor them because you and Alexander asked me to leave the code as he is, because of the size of the code review. If you approve, I refactor this code.

app/controllers/invitations_global_controller.rb
shir: @dissident ok. Leave separate models

app/models/invitation_to_organization.rb
shir: @dissident Yes. Refactor it please.

app/controllers/invitations_global_controller.rb
dissident: @shir I think it is to better name models `Invitation`, `OrganizationInvitation` and `InvitationBase`, what do you think?

shir: @dissident 👍

PR #407, opened 2016-09-28
Stas/missing invite

dissident:
https://trello.com/c/F48VREJH/355-1


app/models/invitation.rb
```diff
@@ -37,4 +43,9 @@ def user_already_in_system
       errors.add(:email, "User already registered in system")
     end
   end
+
+  def week_notification
+    date = 1.week.from_now.beginning_of_day
```
shir: @dissident WHY?!!!!! You may do it much simpler:

- Add scope which will find all unaccepted invitations which were created 1 week ago.
- Via crontab each day resend invitations for each invitation in which scope

That's all. You don't need separate model. You don't need to implement delay logic.

dissident: @shir I agree, but this decision - part of the solution about unsubscriptions. Please see my next PR https://github.com/metastudio/cashbox/pull/409 and you will understand what I mean.

shir: @dissident I saw this PR and still don't understand why do you meed this.

PR #408, opened 2016-10-03
Add organization resources for select fields

yunixon:


app/controllers/api/v1/bank_accounts_controller.rb
```diff
@@ -0,0 +1,15 @@
+module Api::V1
+  class BankAccountsController < ApiController
+
+    api :GET, '/organizations/:organization_id/bank_accounts', 'Return bank accounts for current organization'
+    def index
+      @bank_accounts = current_organization.bank_accounts
+    end
+
+    api :GET, '/organizations/:organization_id/bank_accounts/for_select', 'Return bank accounts for current organization with select format'
+    def for_select
```
shir: @dissident why do you need separate action? It's same as index. If you need data in specified format then it's better to change format on client side, because different selects may use different format and it doesn't make sense to make specific API for each format

PR #411, opened 2016-10-04
Fix disable for transafers, fix typo

yunixon:


app/views/transactions/create_transfer.js.erb
```diff
@@ -23,5 +23,5 @@
     $('#transfer_exchange_rate').parents('#rate_col').removeClass('hidden');
   <% end %>
 <% end %>
-
-addTranferFormMasks();
+$('#submit_btn').prop('disabled', false);
```
AlexanderUlitin: @yunixon how about select button only inside `$("#new_transfer_form")`?

yunixon: @AlexanderUlitin Submit button is not inside form, we use modal and custom #submit_btn for all transactions/transfears forms

PR #417, opened 2016-10-05
Fix bug then on third step of organization wizzard you push back butt…

dissident:
https://trello.com/c/FN2Sl8H3/399-500-wizard


app/controllers/organization_wizzard_controller.rb
```diff
@@ -69,7 +69,7 @@ def require_second_step
     if @organization_wizzard.ready?
       redirect_to root_path
     elsif @organization_wizzard.have_account?
-      redirect_to redirect_to new_category_organization_path
+      redirect_to new_category_organization_path
```
AlexanderUlitin: @dissident can we cover it in specs?

dissident: @AlexanderUlitin I think I can.

PR #422, opened 2016-10-11
Add reset password to API

yunixon:


app/controllers/api/v1/passwords_controller.rb
```diff
@@ -0,0 +1,22 @@
+module Api::V1
+  class PasswordsController < Devise::PasswordsController
+
+    api :POST, '/users/password', 'Reset password'
+    def create
+      @user = User.find_by_email(user_params[:email])
```
shir: @yunixon use `find_by!(email: ...)`. If there is no such user there will be `RecordNotFound` exception which is processed by standard handler as 404 error. Btw `find_by_...` is deprecated.

app/controllers/api/v1/passwords_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+      if @user.present?
+        @user.send_reset_password_instructions
```
shir: @yunixon this will generate email with link to main application. But it should be link to new react application. Also you should add API to change password based on reset password token which you will call from react frontend application

config/config.yml.sample
```diff
@@ -1,5 +1,7 @@
 defaults: &defaults
 
+  edit_password_url: url
```
shir: @yunixon looks like there should be some host with url. It's not obvious from you example. Please make better example.

PR #428, opened 2016-10-12
Fix bug with refresh main page on create transaction

dissident:
https://trello.com/c/oowUe7dB/46--


app/models/main_page_refresher.rb
```diff
@@ -0,0 +1,28 @@
+module MainPageRefresher
```
shir: @dissident looks like it's concern and should be put in `app/models/concern` http://api.rubyonrails.org/classes/ActiveSupport/Concern.html

app/models/main_page_refresher.rb
```diff
@@ -0,0 +1,28 @@
...
+  def prepare_data(transaction)
+    organization = transaction.organization
+    transaction.bank_account.touch
```
shir: @dissident not sure why do you make `touch` here. You don't make any data update here but touch is required only for update.

app/models/main_page_refresher.rb
```diff
@@ -0,0 +1,28 @@
...
+    {
+      id: "#transaction_#{transaction.id}",
```
shir: @dissident use [`dom_id`](http://apidock.com/rails/ActionView/RecordIdentifier/dom_id)

app/models/main_page_refresher.rb
dissident: @shir our model bank_account has `after_touch :recalculate_amount!`, but `touch` work only associated model (`transaction` in current example) change `updated_at`, it's not working after :create or :destroy. I think we can do new task for this bug, because bank_account must recalculate amount after create and destroy of transactions.

dissident: http://guides.rubyonrails.org/association_basics.html#touch

dissident: @shir `touch` not working after create transaction

dissident: @shir we can do new task for fix this bug, or I can fix it in current task.

shir: @dissident fix in current. And it's not a bug. It's just wrong written code. In this method you dont make any save so nothing to recalculate. If you make any updated in other method and then call this method then it's wrong. This touch should be where update is done.

Fix it here.

PR #433, opened 2016-10-18
Fix customers API

yunixon:


app/controllers/api/v1/customers_controller.rb
```diff
@@ -41,7 +41,8 @@ def update
 
     api :DELETE, '/organizations/:organization_id/customers/:id', 'Destroy customer'
     def destroy
-      @customer.destroy!
+      @customer.destroy
+      render :show # acts_as_paranoid
```
shir: @yunixon what does this comment mean?

PR #435, opened 2016-10-18
Fix categories api

yunixon:


app/controllers/api/v1/categories_controller.rb
```diff
@@ -41,7 +41,8 @@ def update
 
     api :DELETE, '/organizations/:organization_id/categories/:id', 'Destroy category'
     def destroy
-      @category.destroy!
+      @category.destroy
+      render :show # acts_as_paranoid
```
shir: @yunixon what does this comment mean?

PR #451, opened 2017-02-01
Set current date in the copied transaction date field

psmir:
https://trello.com/c/WmmnYzwi/436--


app/controllers/transactions_controller.rb
```diff
@@ -22,6 +22,7 @@ def new
       else
         @transaction = @transaction.dup
         @transaction.amount = @transaction.amount.abs
+        @transaction.date = DateTime.now
```
AlexanderUlitin: @psmir it's better to use `Time.current`, cos it uses current timezone, instead of server timezone.

PR #452, opened 2017-02-02
Change ransackable scope on where, first not worked

dissident:


dissident:
[#413 NoMethodError: undefined method `[]' for nil:NilClass](https://trello.com/c/IVWZeYKn/437-413-nomethoderror-undefined-method-for-nil-nilclass)

PR #463, opened 2018-03-06
Add api for user profile and account management

dissident:


app/controllers/api/v1/users_controller.rb
```diff
@@ -1,9 +1,85 @@
...
 
+      api :DELETE, '/users/:id', 'Destroy User with this id if its you'
+      def destroy
+        @user.destroy
+      end
+
+      private
+
+      def update_profile_params
+        params.require(:user).permit(
```
shir: @dissident instead of `require(:user)` it's better to use `fetch(:user, {})` in API controllers. Because in api if there is no `user` param in request first variant will throw error 500, but second will return 422 which describe which fields are missing

app/policies/user_policy.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def destroy?
+    record_is_member?
+  end
+
+  private
+
+  def record_is_member?
+    @record == @member
```
shir: @dissident it's better to compare by `id`

PR #475, opened 2018-04-27
Add download pdf to show action

dissident:


app/views/api/v1/invoices/_invoice_body.html.slim
```diff
@@ -0,0 +1,56 @@
+table.table.invoice class="#{'table-bordered' if request.format.html?}"
```
shir: @dissident don't duplicate view. Use existed one.

PR #480, opened 2018-05-10
update members and transactions api

yunixon:
https://trello.com/c/fty57MFm/476-transactions-%D0%BF%D0%BE%D0%B4%D1%81%D0%B2%D0%B5%D1%87%D0%B8%D0%B2%D0%B0%D0%B5%D1%82-%D0%BD%D0%BE%D0%B2%D1%8B%D0%B5-%D1%82%D1%80%D0%B0%D0%BD%D0%B7%D0%B0%D0%BA%D1%86%D0%B8%D0%B8


app/controllers/api/v1/members_controller.rb
```diff
@@ -26,14 +26,32 @@ def update
...
+    api :GET, '/organizations/:organization_id/members/:id', 'Return member'
+    def show
+    end
+
+    api :GET, '/organizations/:organization_id/member_info', 'Return current member'
+    def current
+      @member = current_member
+    end
+
+    api :PUT, '/organizations/:organization_id/members/:id/update_last_viewed_at', 'Update last viewed at'
```
shir: @yunixon there is no reasons to include `:id` in url, because user may update only himself `last_viewed_at`. I think you should add separate url to work with current member only which will not include `:id` (may be `current_member`?). And `update_last_viewed_at` is bad name for path (but not for action). Path should not include verb word. May be PUT `last_visit` will work. Not sure.

app/policies/application_policy.rb
```diff
@@ -53,7 +53,8 @@ def resolve
   end
 
   protected
-    def owner_or_admin_with_access?
-      (member.admin? && record.role != 'owner') || member.owner?
-    end
+
+  def owner_or_admin_with_access?
+    (member.admin? && record.role != 'owner') || member.owner?
```
shir: @yunixon why you check `role` for `record`? `record` may be any model and may have not `role` field.

app/controllers/api/v1/members_controller.rb
```diff
@@ -26,14 +26,32 @@ def update
...
+    def show
+    end
+
+    api :GET, '/organizations/:organization_id/member_info', 'Return current member'
+    def current
+      @member = current_member
+    end
+
+    api :PUT, '/organizations/:organization_id/last_visit', 'Update last viewed at for current member'
+    def update_last_viewed_at
```
shir: @yunixon rename method to `update_last_visit` so it will match route name

app/controllers/api/v1/members_controller.rb
```diff
@@ -26,14 +26,32 @@ def update
...
+      if current_member.update(last_visited_at: Time.current)
```
shir: @yunixon add ability to pass `last_visited_at` as param, but if it's not passed, then use `Time.current`

app/models/transaction.rb
```diff
@@ -158,6 +158,11 @@ def get_type
     end
   end
 
+  def viewed_for_member?(member)
+    return false if member&.last_visited_at.blank?
```
shir: @yunixon return `true` if `last_visited_at` blank. For first visit it's better to show all transactions as viewed.

PR #483, opened 2018-05-17
fix broken specs, warnings and deprecations

yunixon:
https://trello.com/c/4f4ZEGIq/439-fix-failed-specs-deprecations

### DIFF
--- a/spec/requests/api/v1/bank_accounts/index_spec.rb
+++ b/spec/requests/api/v1/bank_accounts/index_spec.rb
@@ -18,19 +18,23 @@ describe 'GET /api/organizations/#/bank_accounts' do
     it 'returns bank_accounts' do
       expect(response).to be_success
 
-      expect(json[0]).to include(
-        'id'       => bank_account2.id,
-        'name'     => bank_account2.name,
-        'currency' => bank_account2.currency,
-        'balance'  => bank_account2.balance.as_json,
-        'residue'  => bank_account2.residue.as_json,
+      expect(json).to include(
+        'id'              => bank_account2.id,
+        'name'            => bank_account2.name,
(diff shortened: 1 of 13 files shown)


spec/requests/api/v1/bank_accounts/index_spec.rb
```diff
@@ -18,19 +18,23 @@
     it 'returns bank_accounts' do
       expect(response).to be_success
 
-      expect(json[0]).to include(
-        'id'       => bank_account2.id,
-        'name'     => bank_account2.name,
-        'currency' => bank_account2.currency,
-        'balance'  => bank_account2.balance.as_json,
-        'residue'  => bank_account2.residue.as_json,
+      expect(json).to include(
```
shir: @yunixon not sure why do you need this fix. `json` should have array of bank_accounts. Why you change to compare `json` instead `json[0]`?

spec/support/capybara.rb
```diff
@@ -4,9 +4,13 @@
...
 
-# Capybara::Webkit.configure do |config|
-#   config.skip_image_loading
-# end
+Capybara::Webkit.configure do |config|
+  # Silently return an empty 200 response for any requests to unknown URLs.
+  config.block_unknown_urls
+
+  # Timeout if requests take longer than 20 seconds
+  config.timeout = 20
```
shir: @yunixon this timeout looks to long for tests. Change it to 5 seconds.

yunixon: @shir with 5 seconds some specs randomly failed

shir: @yunixon it's the issue you should fix. No requests should take more then 5 seconds.

PR #479, opened 2018-05-08
Fix deploy recipes

AlexanderUlitin:
Updated deploy configs for new production server

PR #486, opened 2018-05-28
Fix deploy configs

AlexanderUlitin:
Updated configs for correct deploy to production


AlexanderUlitin:
@shir wrong PR, cos I did the same in https://github.com/metastudio/cashbox/pull/479

PR #493, opened 2018-06-04
fix nested errors, update invoices and users api

yunixon:
https://trello.com/c/Rs36Jktk/465-invoices-edit-edit-items

### DIFF
--- a/lib/nested_errors.rb
+++ b/lib/nested_errors.rb
@@ -5,7 +5,7 @@ module NestedErrors
     errors = object.errors.messages.to_hash
     return errors if relationships.blank?
 
-    relationships.each do |relationship|
+    [*relationships].each do |relationship|
       errors[relationship] = {}
       association = object.__send__(relationship)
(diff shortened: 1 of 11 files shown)


lib/nested_errors.rb
```diff
@@ -1,10 +1,27 @@
...
-    hash.each_with_object({}) do |(key, value), all|
-      regex = key.to_s.include?('[') ? /\[|\]\./ : '.'
-      key_parts = key.to_s.split(regex).map!(&:to_sym)
-      leaf = key_parts[0...-1].inject(all) { |h, k| h[k] ||= {} }
-      leaf[key_parts.last] = value
+  def self.unflatten(object, relationships)
+    errors = object.errors.messages.to_hash
+    return errors if relationships.blank?
+
+    relationships.each do |relationship|
```
shir: @yunixon you may use `[*relationships]` this will allow to pass `:invoice_items` instead `[:invoice_items]` for single element, which is more readable.

PR #494, opened 2018-06-20
Add api for debtors sidebar

dissident:


dissident:
DO NOT MERGE!!!!! DANGER!!!!!!

yunixon:
@dissident Please add specs for debtors api

app/controllers/api/v1/base_controller.rb
```diff
@@ -35,6 +35,22 @@ def currencies
       render json: Dictionaries.currencies
     end
 
+    def currency_rates
```
shir: @dissident why it's in api base controller? There should be no any actions in base controller.

app/controllers/api/v1/base_controller.rb
```diff
@@ -35,6 +35,22 @@ def currencies
...
+      bank = Money.default_bank
+      rates = bank.rates.inject([]) do |result, (k, v)|
```
shir: @dissident you should move it separate model or services or... anything. But this logic should not be in controller.

app/controllers/api/v1/base_controller.rb
shir: @dissident not sure why we ever need this action. All currency calculating should be done on backend side.

app/models/customer.rb
```diff
@@ -26,14 +26,21 @@ class Customer < ApplicationRecord
...
   # gem 'paranoia' doesn't run validation callbacks on restore
   before_restore :run_validations
 
   def to_s
     name.truncate(30)
   end
 
+  def indebtedness
+    result = { 'EUR': 0, 'USD': 0, 'RUB': 0 }
+    invoices.unpaid.pluck(:currency, :amount_cents).map do |i|
```
shir: @dissident instead calculating in ruby summarize everything via sql using `group by`

app/models/customer.rb
```diff
@@ -26,14 +26,21 @@ class Customer < ApplicationRecord
...
-
   # gem 'paranoia' doesn't run validation callbacks on restore
   before_restore :run_validations
 
   def to_s
     name.truncate(30)
   end
 
+  def indebtedness
+    result = { 'EUR': 0, 'USD': 0, 'RUB': 0 }
```
shir: @dissident not sure it's good to use predefined list of currencies. In future we will have a lot of currencies, so when we will add new currency, we will have to edit this code (and almost always will forget to do it). Instead I suggest to make as it is in current logic: have separate row for each customer-currency pair. In real case each customer will have only one currency.

app/controllers/api/v1/base_controller.rb
shir: @dissident why this code is still here?

app/controllers/api/v1/debtors_controller.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+module Api::V1
+  class DebtorsController < BaseOrganizationController
+    def index
+      debtors = []
```
shir: @dissident move this all to model or separate class (presenter?)

app/controllers/api/v1/debtors_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+          debtors.push({
+            id: debtor.id,
+            name: debtor.name,
+            amount: amount
+          })
+          total += amount[:total]
+        end
+        indebtedness_summ!(summ_by_currencies, indebtedness)
+      end
+      render json: {
```
shir: @dissident move it to views

app/controllers/api/v1/debtors_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+        debtors: debtors,
+        summ_by_currencies: summ_by_currencies,
+        total: total
+      }.to_json
```
shir: @dissident if you render hash as json you don't have to convert it to json.

app/controllers/api/v1/debtors_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    private
+
+    def indebtedness_summ!(summ, indebtedness)
```
shir: @dissident move this too

app/controllers/api/v1/debtors_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+      indebtedness.each do |debt|
+        name = "All customers (#{debt.currency})"
+        search = summ.select { |d| d[:name] == name }.first
```
shir: @dissident why do you use `select{ ... }.first`? There is a method [`find`](https://apidock.com/ruby/Enumerable/find)

app/controllers/api/v1/debtors_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+        debtors: debtors,
+        summ_by_currencies: summ_by_currencies,
+        total: total
+      }.to_json
+    end
+
+    private
+
+    def indebtedness_summ!(summ, indebtedness)
+      indebtedness.each do |debt|
```
shir: @dissident and why can't you do it using sql `SUM` and `GROUP BY`? Check out `invoices_debt` method in `layout_helper.rb` which makes same calculations.

app/models/converted_money_presenter.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
```
shir: @dissident move presenter to `app/presenters` folder

app/models/converted_money_presenter.rb
```diff
@@ -0,0 +1,48 @@
...
+    @old_amount = money
+    @new_currency = new_currency
+  end
+
+  def present
+    return not_converted if @old_amount.currency == @new_currency
+    converted
+  end
+
+  def converted
```
shir: @dissident instead making two method which returns hash with same structure I suggest to make one method which will call methods for each field which return values and this field methods this return values dependent to conditions.

app/models/customer.rb
```diff
@@ -26,14 +26,25 @@ class Customer < ApplicationRecord
...
   # gem 'paranoia' doesn't run validation callbacks on restore
   before_restore :run_validations
 
   def to_s
     name.truncate(30)
   end
 
+  def indebtedness
+    result = {}
+    invoices.unpaid.pluck(:currency, :amount_cents).map do |i|
```
shir: @dissident use sql `SUM` with `GROUP BY`

app/models/converted_money_presenter.rb
```diff
@@ -0,0 +1,48 @@
...
+  def new_amount
+    @old_amount.exchange_to(@new_currency)
+  end
+
+  def rate
+    Money.default_bank.get_rate(@old_amount.currency, @new_currency)
+  end
+
+  def updated_at
+    I18n.l Money.default_bank.rates_updated_on
```
shir: @dissident return dates as is. They will be converted in json by rails. Don't format them on backend.

app/models/debts/debt.rb
```diff
@@ -0,0 +1,36 @@
...
+        m = Money.new(amount_cents, currency)
+        return m + total if def_curr == currency
+        m.exchange_to(def_curr) + total
+      end
+    end
+
+    private
+
+    def invoices_group_by_currency
+      return @invoices_group_by_currency if @invoices_group_by_currency.present?
```
shir: @dissident why can't you use `@invoices_group_by_currency ||= ...`?

app/models/debts/debt.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module Debts
+  class Debt
+    def initialize(instance)
```
shir: @dissident `@instance` is a bad name for variable. What does it mean? Instance of that does it handle?

app/models/debts/base_debt.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    def by_currency
+      invoices_group_by_currency.map do |currency, amount_cents|
+        Money.new(amount_cents, currency)
+      end
+    end
+
+    def total
+      def_curr = @organization.default_currency
```
shir: @dissident pass `@organization` through initializer or use "virtual" method. Don't use class variable which are defined in child classes.

app/models/debts/base_debt.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module Debts
+  # don't use this class directly, use OrganizationDebt or CustomerDebt
+  class BaseDebt
```
shir: @dissident may be it will be better to use concern instead base class?

spec/presenters/converted_money_presenter_spec.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
```
shir: @dissident I think we don't need this spec. This presenter will be tested when you test `debtors/index`

PR #647, opened 2024-10-04
grigorii/export invoices and transaction to csv

MartGS:
Add function export invoices & transactions to csv format

### DIFF
--- a/app/models/concerns/generate_csv.rb
+++ b/app/models/concerns/generate_csv.rb
@@ -1,0 +1,26 @@
+module GenerateCsv
+  extend ActiveSupport::Concern
+  require 'csv'
+
+  class_methods do
+    def to_csv
+      CSV.generate(col_sep: ';') do |csv|
+        csv << column_names_for_export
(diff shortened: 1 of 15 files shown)


app/models/concerns/generate_csv.rb
```diff
@@ -0,0 +1,26 @@
+module GenerateCsv
+  extend ActiveSupport::Concern
+  require 'csv'
+
+  class_methods do
+    def to_csv
+      CSV.generate(col_sep: ';') do |csv|
+        csv << column_names_for_export
+        order(created_at: :desc).each do |item|
```
AlexanderUlitin, line 9: @MartGS  It's better to use `find_each` method (more optimal for long lists and it does not return the full list of objects as result).

AlexanderUlitin, line 9: @MartGS let's remove `order(created_at: :desc)` from this method, so we will use order of the collection that was applied above of it.

AlexanderUlitin (changes requested):

